PR #5, opened 2023-04-23
feat: add request-transaction waku object

sugarpusher:


src/lib/components/button.svelte
```diff
@@ -18,18 +28,42 @@
...
+		}
+
+		&.right {
+			margin-left: auto;
+			margin-right: 0;
+		}
+
+		&.reverse {
+			flex-direction: row-reverse;
+		}
```
vojtechsimetka, line 61: Is this because of having icon on the right side of button? I would consider instead to have two icon properties instead `iconStart` and `iconEnd`. This is more consistent with UI frameworks.

src/lib/components/container.svelte
```diff
@@ -1,24 +1,38 @@
 <script lang="ts">
 	export let direction: 'row' | 'column' = 'column'
 	export let gap = 0
+	export let variant: 'space-between' | 'flex-start' | 'flex-end' | 'center' = 'flex-start'
```
vojtechsimetka, line 4: I think we could name this `justify` instead of variant.

src/lib/components/container.svelte
```diff
@@ -1,24 +1,38 @@
...
 		flex-direction: row;
 	}
 	.column {
 		flex-direction: column;
 	}
+	.space-between,
+	.stretch,
+	.center {
+		height: 100%;
+	}
```
vojtechsimetka, line 28: Why `flex-start` and `flex-end` don't have 100% height? It actually seems that these options are ignored and instead there is `stretch`

src/lib/components/icons/checkmark.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+	import type { IconProps } from '$lib/types'
+
+	type $$Props = IconProps
+
+	let cls: string | undefined = undefined
+	export { cls as class }
```
vojtechsimetka, line 7: Just so you know, we should remove these. Adding class to a component is now done differently (and we don't need it)

src/lib/components/icons/edit.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+	import type { IconProps } from '$lib/types'
+
+	type $$Props = IconProps
+
+	let cls: string | undefined = undefined
+	export { cls as class }
```
vojtechsimetka, line 7: Just so you know, we should remove these. Adding class to a component is now done differently (and we don't need it)

src/routes/+layout.svelte
```diff
@@ -31,12 +33,17 @@
 		height: 100dvh;
 		width: 100vw;
 		background-color: black;
+		/* padding: 0 var(--spacing-24); */
 	}
 
 	.wrapper {
-		max-height: 915px;
+		/* max-height: calc(100vh - var(--spacing-24)); */
```
vojtechsimetka, line 40: Why did you decide to do this? :) I am OK with this altough I quite liked having it look a bit more like a phone screen (this was Samsung Galaxy S20 Ultra)

vojtechsimetka (changes requested):
I noticed the start chat screen got broken a bit.
Screenshot 2023-04-27 at 7 50 06

src/lib/components/button.svelte
vojtechsimetka: Here is an example from Material UI (React) https://mui.com/material-ui/react-button/#buttons-with-icons-and-label

Or we can (and this may be better) change it from `props` to `slot`.

sugarpusher: I haven't seen a design for this page so I'm not sure what it's supposed to look like. What are you using as reference? Can you send me a link?

src/lib/components/container.svelte
sugarpusher: Agreed

src/lib/components/icons/checkmark.svelte
sugarpusher: Oh cool, good to know! I'll remove it.

src/lib/components/icons/edit.svelte
sugarpusher: 👍

src/routes/+layout.svelte
sugarpusher: I think that limiting the height to a specific size puts us in a situation where it may not look right on different (taller) devices, and there's no reason for that. The .root is limiting the height to 100vh so that makes sure we're constrained to the device height.
But also probably because since I'm working on a desktop, I want to be able to have as much viewing area as possible. 😉

src/lib/components/button.svelte
vojtechsimetka: I assumed you add this so that you can have the icon after the label, right? If my assumption is wrong then please explain why we have this :)

sugarpusher: Your assumption is correct 😁

src/routes/+layout.svelte
sugarpusher: Oh the scrolling works because I have a super long chat with myself going to test that 😅
And the max-height line is currently commented out but I think when I added it I had some padding in the .root so I was making up for it.

src/lib/components/button.svelte
vojtechsimetka: OK then I would prefer to either use slots for the icons or props.

```html
<Button>
   <Icon slot='start' />
    label
   <Icon slot='end' />
</Button>

<!-- OR -->
<Button iconStart={Icon} iconEnd={Icon}>
    label
</Button>
```

I have slight preference to use the slot variant

src/lib/components/container.svelte
vojtechsimetka: I just pointed this out for consistency reasons.

src/lib/components/button.svelte
sugarpusher: Yup we can do that, let's chat so I understand the reason for the preference so I can make more informed decisions moving forward 😁

src/lib/components/container.svelte
sugarpusher: Checking this out now, it looks like adding 100% to flex-start and flex-end does cause issues but I think what I need to do is restructure some other elements/components so I'll do that. 😉

PR #71, opened 2023-05-24
style: new chat flow

sugarpusher:


src/lib/components/container.svelte
```diff
@@ -4,10 +4,11 @@
 	export let justify: 'space-between' | 'flex-start' | 'flex-end' | 'center' = 'flex-start'
 	export let align: 'left' | 'right' | 'center' | undefined = undefined
 	export let grow: boolean | undefined = undefined
+	export let sticky: 'top' | 'bottom' | 'none' = 'none'
```
vojtechsimetka, line 7: We should use everywhere either `undefined` or `'none'` everywhere.

src/routes/+page.svelte
```diff
@@ -139,7 +139,12 @@
 										{formatDateAndTime(lastMessage.timestamp)}
 									</span>
 								</div>
-								<p class="message text-serif">
+								<p
+									class={`message text-serif ${
+										lastMessage.fromAddress === $profile.address ? 'my-message' : ''
+									}`}
+								>
+									{lastMessage.fromAddress === $profile.address ? 'You: ' : ''}
```
vojtechsimetka, line 147: We should extract this check into svelte constant `{@const myMessage=lastMessage.fromAddress === $profile.address}`

vojtechsimetka (approved):

PR #72, opened 2023-05-25
feat: waku integration

agazso:


vojtechsimetka (approved):
I'd suggest we at least for now allow switching between the adapters like we did in Kurate

https://kurate.vercel.app/dev

https://github.com/logos-innovation-lab/Kurate/blob/main/packages/ui/src/lib/adapters/index.ts#L38-L60

https://github.com/logos-innovation-lab/Kurate/blob/main/packages/ui/src/routes/dev/%2Bpage.svelte

PR #93, opened 2023-06-05
feat: add token balances to the app

vojtechsimetka:
implements #92


src/lib/components/asset.svelte
```diff
@@ -1,20 +1,22 @@
...
 		<div class="token">
-			<img src="https://picsum.photos/48/48" alt={`${name} logo`} />
+			<img src={image} alt={`${name} logo`} />
 			<span class="text-lg text-bold">
 				{name}
 			</span>
 		</div>
 		<div class="text-lg">
-			{amount}
+			{(Number(amount) / 10 ** decimals).toFixed(2)}
```
filoozom, line 19: I'm not sure whether there can actually be a precision difference as the lower bits are all discarded, but if there is, this could potentially be more precise with:
```js
Number(amount / 10n ** BigInt(decimals - 2)) / 100
```

src/lib/stores/wallet.ts
```diff
@@ -4,8 +4,17 @@ import { getFromLocalStorage, removeFromLocalStorage, saveToLocalStorage } from
...
+	name: string
+	symbol: string
+	amount: bigint
+	decimals: number
+	image: string
+}
+
 export interface WalletState {
 	wallet?: HDNodeWallet
+	balances?: Token[]
```
agazso, line 17: Maybe it's better not mixing optionals with arrays, because the empty array can also express optionality, but the two cases has to be handled differently in the code.

PR #97, opened 2023-06-09
feat: waku object hello world example

agazso:
Current limitations:
- [x] every data message displays a WakuObject
    - per the discussion it is okay for now
- [x] ~only Firebase backend is supported but not the Waku~
- [x] in the Firebase adapter only the latest data message is checked state update
    - per the discussion it is okay for now
- [x] the waku object store is generic and therefore not type safe
    - per the discussion it is okay for now
- [x] there is a missing mechanism for updating the object store state from the message (e.g. a reducer function)
    - per the discussion it is okay for now


src/lib/objects/hello-world/hello-world-receiver.svelte
```diff
@@ -0,0 +1,35 @@
...
+	import Button from '$lib/components/button.svelte'
+	import type { DataMessage } from '$lib/stores/chat'
+	import { objectKey, objectStore } from '$lib/stores/objects'
+	import { profile } from '$lib/stores/profile'
+	import { walletStore } from '$lib/stores/wallet'
+	export let message: DataMessage
+
+	$: name = (
+		$objectStore.objects.get(objectKey(message.objectId, message.instanceId)) as { name?: string }
+	)?.name
```
vojtechsimetka, line 13: On a call we talked about passing these things in as props. I am fine to merge it as is, just want to make sure you have not forgot about it and this is here left deliberately.

vojtechsimetka (approved):

src/lib/objects/hello-world/hello-world-receiver.svelte
agazso: Indeed, I will merge this and then I will work on passing props as the next step.

PR #100, opened 2023-06-12
feat: send transaction waku object wip

vojtechsimetka:
**TODO:**
- [ ] overview view
- [ ] detect who is the sender and who is recipient
- [x] object schema validation
- [ ] await on transaction finish
- [x] make the `chats/[id]/object/[object_id]/new/[[view]]` page universal
- [ ] implement fiat pricing
- [ ] validate user has enough money

**Not part of this PR:**
- waku & blockchain implementation


agazso:
I get an error after the transaction seemingly successfully executed:

screenshot

src/lib/objects/index.d.ts
```diff
@@ -9,11 +10,15 @@ export interface WakuObjectArgs {
...
 	send: (data: unknown) => Promise<void>
+	sendTransaction?: (to: string, token: Token, fee: Token) => Promise<string>
+	estimateTransaction?: (to: string, token: Token) => Promise<Token>
+	onViewChange?: (view: string) => void
 }
 
 interface WakuObjectDescriptor {
 	readonly objectId: string
 	readonly wakuObject: ComponentType
+	readonly standalone?: ComponentType
```
agazso, line 21: I assume this is what is shown when the object is being set up.

If yes, I would like to find a better name for it later. I think there will be three different major views (components) for an object:
1. when you set it up
2. chat widget
3. info screen

From the current designs 1. and 2. can be internally represented with other views (components).

src/lib/objects/index.d.ts
```diff
@@ -9,11 +10,15 @@ export interface WakuObjectArgs {
 	updateStore: (updater: (state: unknown) => unknown) => void
 
 	send: (data: unknown) => Promise<void>
+	sendTransaction?: (to: string, token: Token, fee: Token) => Promise<string>
+	estimateTransaction?: (to: string, token: Token) => Promise<Token>
+	onViewChange?: (view: string) => void
```
agazso, line 15: I guess this is called when the view changes and at one point this is called with the argument `overview`. As mentioned in another comment it would be good to settle for the naming later and be consistent with it.

agazso (approved):
Looks good in general, however it would be good to know what is the cause of the error before merging.

src/lib/objects/index.d.ts
vojtechsimetka: I don't think the naming can be standardised. We can not enforce the amount of views nor their names to the waku object devs. However, I don't really thing we care as long as it is URL safe string.

src/lib/objects/index.d.ts
vojtechsimetka: yes that is the case. Also it can be used to show `3. info screen` but likely that should be a different component.

vojtechsimetka:
So the reason for the error is that `message.data` may not be populated yet. Not sure really why, but for now it is fixed with https://github.com/logos-innovation-lab/waku-objects-playground/pull/100/commits/4614ab6c3829946f34378cd04a69670090a2c682#diff-682f0f3dcd82eab92c405c874027d2e250f3330f2d37d752e81444f3f533d3aaR34-R36

PR #104, opened 2023-06-14
feat: waku object interface

agazso:
First iteration of the waku object interface (see `WakuObjectArgs`) for communicating between the host and the object. Currently limited to `name`, `address`, the object instance's own `store` and a way to `send` data messages in the chat context to the peer instance.

Later this can be extended with more information about the user or with extra functionality (e.g. wallet). Also this can serve as the basis for using it with `postMessage` in restricted environments.

The `unknown`s are quite ugly, it would be great if we could find a better, generic solution to be able to avoid casting. One idea that came up already was using validation with `zod`.


vojtechsimetka (approved):
I'll think some more about how to do the types. But from my point of view as of right now the `unknown` is the best we can do.

PR #105, opened 2023-06-15
feat: waku object updateStore

agazso:
This PR adds the functionality to update the object store in a generic way when an external event happens. Currently it's only updated when there is a data message coming in, but later more events (e.g. transactions) can be added.

The idea is that each object defines a `WakuObjectDescriptor` which may define functionality what should happen when an external event happens (e.g. `onMessage`). I realized that in order to keep the data flow uni-directional the update needs to happen through the adapter and that will update the store. I attached a small diagram about this (which require more work I admit :smile:  ).

Known issues: 
- the firebase update checks the `timestamp` of a message to see if it has been applied for the store already. However this may introduce inconsistencies. For example if there are two events happening in about the same time on different nodes then when merging local updates may be visible and then remote updates may be ignored.
- the firebase update reads the current state of the object store to pass it as an argument to the `updater` function and then using the return value to update the Firebase store and then the objectstore will be updated in the callback of the subscription of the Firebase store. This makes it difficult to follow the code and most likely is suboptimal.

Untitled Diagram drawio

### DIFF
--- a/src/lib/adapters/firebase/index.ts
+++ b/src/lib/adapters/firebase/index.ts
@@ -33,10 +33,11 @@ import { formatAddress } from '$lib/utils/format'
 import { db, ipfs, IPFS_GATEWAY } from './connections'
 
 import type { HDNodeWallet } from 'ethers'
-import type { Unsubscriber } from 'svelte/store'
+import { get, type Unsubscriber } from 'svelte/store'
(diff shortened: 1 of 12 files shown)


vojtechsimetka (approved):
Works well, thank you!

PR #106, opened 2023-06-16
style: add chat bubble to Hello World messages

sugarpusher:


src/lib/objects/waku-object.svelte
```diff
@@ -45,7 +45,9 @@
 	}`}
 >
 	<div class="message-content">
-		<svelte:component this={component} {message} {args} />
+		<div class="message-text text-lg">
```
vojtechsimetka, line 48: does it need to be wrapped in two `div`s or can you just add the classes to the outer `div`?

sugarpusher: It does work with only one div. I think the reason why I started using separate divs is to give us more  flexibility when styling more complex objects - so I was just following the initial pattern. But looking at my test objects it looks like I can consolidate those as well. I'll do that.
Thanks!

agazso (approved):

vojtechsimetka (approved):

PR #107, opened 2023-06-17
fix: show other person's avatar on chat

sugarpusher:


src/routes/chat/[id]/+page.svelte
```diff
@@ -103,7 +103,14 @@
...
 
-	let chatImg = $chats.chats.get($page.params.id)?.users[0].avatar
+	let chatImg: string | undefined = undefined
+	if ($chats.chats.get($page.params.id)?.users.length === 2) {
+		$chats.chats.get($page.params.id)?.users.forEach((member) => {
+			if (member.address !== $walletStore.wallet?.address) {
+				chatImg = member.avatar
+			}
+		})
+	}
```
Comment (vojtechsimetka) [inline src/routes/chat/[id]/+page.svelte:113]: Thanks, this is quite correct. However, just FYI I would write it this way (not tested)

```suggestion
let chatImg: string | undefined = undefined
$: if ($chats.chats.get($page.params.id)?.users.length === 2) {
    const otherUser = $chats.chats.get($page.params.id)?.users.find(m => m.address !== $walletStore.wallet?.address)
    chatImg = otherUser?.avatar
}
```

Comment (vojtechsimetka) [reply src/routes/chat/[id]/+page.svelte:113]: You can also to it all in one line but that would be really hard to read (even this is).

Comment (vojtechsimetka) [reply src/routes/chat/[id]/+page.svelte:113]: Also it may be useful to actually have the `otherUser` be available because in addition to the chat image being `otherUser.avatar`, the name of the chat also is  `otherUser.name`.

Comment (sugarpusher) [reply src/routes/chat/[id]/+page.svelte]: @vojtechsimetka 
Cool, thank you! I applied your suggestion manually because I wanted to test first 😊
Yes, I think initially we were typing in the chat name but if we're only doing 1-1 chats for now, we should use otherUser. I'll look into that too.
Thanks!

src/routes/chat/[id]/+page.svelte
```diff
@@ -103,7 +103,13 @@
...
 	}
 
-	let chatImg = $chats.chats.get($page.params.id)?.users[0].avatar
+	let chatImg: string | undefined = undefined
+	const otherUser = $chats.chats
+		.get($page.params.id)
+		?.users.find((m) => m.address !== $walletStore.wallet?.address)
+	$: if ($chats.chats.get($page.params.id)?.users.length === 2) {
+		chatImg = otherUser?.avatar
+	}
```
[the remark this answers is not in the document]
Comment (vojtechsimetka) [reply src/routes/chat/[id]/+page.svelte:112]: You have to do it this way, you can't use const for `otherUser` because it gets populated when the component is mounted and likely will be undefined.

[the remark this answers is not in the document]
Comment (sugarpusher) [reply src/routes/chat/[id]/+page.svelte:112]: Yup that's what I thought. Thank you!

vojtechsimetka (approved):

Issue #103, opened 2023-06-13
The message input on the chat screen moves when text is entered


agazso:
Both on Firefox and Chrome

Empty
![Screenshot from 2023-06-13 15-58-54](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/d5f02908-f34c-43ae-8ae5-963ec96a240d)

With text
![Screenshot from 2023-06-13 15-59-01](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/0e49f303-750e-46a0-845c-eec2ef76a922)

agazso:
Closed by #109

Issue #102, opened 2023-06-13
The share link is misaligned in the Invite screen on Firefox


agazso:
![Screenshot from 2023-06-13 15-54-58](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/bd73ec22-19da-4808-8977-1c6eeb6b9de0)

agazso:
Closed by #109

Issue #101, opened 2023-06-13
Display name issues in the Create New Identity screen


agazso:
I found a few issues with the Display name input field in the Create New Identity screen.

- Pressing Enter creates a new line. It would be enough to have a single line input field. Maybe it can also accept the value and press the button to go to the next screen. [1]
- The Display name field is displayed incorrectly on Firefox when it's empty. After the first character is entered it is displayed correctly. [2]
- Maybe it would be better to focus on the Display name input field. If not, then a placeholder would be good.

[1]
![Screenshot from 2023-06-13 15-48-51](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/0cf9763d-bb49-425d-bc66-5fd4b02d6275)

[2]
![Screenshot from 2023-06-13 15-41-30](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/3e0fc9e7-5bb6-4395-94f2-bc485ed3a485)

agazso:
Closed by #109

PR #109, opened 2023-06-20
fix: fix textarea component bugs

sugarpusher:
Also adds Input Field component

Issues addressed in this PR
#101
#102
#103

### DIFF
--- a/src/lib/components/textarea.svelte
+++ b/src/lib/components/textarea.svelte
@@ -6,6 +6,7 @@
 	export let placeholder = ''
 	export let autofocus = false
 	export let disabled = false
+	export let readonly = false
 	export let pad = 0
 	export let height = 46
 
@@ -64,6 +65,7 @@
 			{disabled}
 			{autofocus}
 			{placeholder}
+			{readonly}
 			rows="1"
 		/>
 	</div>
@@ -129,7 +131,8 @@
 			color: var(--black);
 		}
 
-		&:disabled {
+		&:disabled,
+		&:read-only {
 			background-color: var(--gray10);
 			border-color: transparent;
 		}
--- a/src/lib/components/textarea.svelte
+++ b/src/lib/components/textarea.svelte
@@ -7,16 +7,16 @@
 	export let autofocus = false
 	export let disabled = false
 	export let pad = 0
-	export let rows = 1
+	export let height = 46
 
 	let placeholderHeight: number
 	let textarea: HTMLTextAreaElement
 
 	const resizeEvents = ['change']
-	const delayedResizeEvents = ['cut', 'paste', 'drop', 'keydown']
+	const delayedResizeEvents = ['change', 'cut', 'paste', 'drop', 'keydown']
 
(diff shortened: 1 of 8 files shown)


src/routes/identity/account/+page.svelte
```diff
@@ -48,7 +48,7 @@
 		<p class="text-lg">This address is used to send and receive tokens with Waku objects</p>
 	</Container>
 	<Container gap={6}>
-		<Textarea value={$walletStore.wallet?.address} disabled label="Account address" rows={2} />
+		<Textarea value={$walletStore.wallet?.address} disabled label="Account address" />
```
vojtechsimetka, line 51: These likely should not be Textareas but some other components. The reason is that you can not select and copy from disabled textarea.

sugarpusher: 100% agree.

src/lib/components/textarea.svelte
```diff
@@ -6,17 +6,18 @@
...
 	let placeholderHeight: number
 	let textarea: HTMLTextAreaElement
 
 	const resizeEvents = ['change']
-	const delayedResizeEvents = ['cut', 'paste', 'drop', 'keydown']
+	const delayedResizeEvents = ['change', 'cut', 'paste', 'drop', 'keydown']
 
 	function resize() {
-		textarea.style.height = 'auto'
+		textarea.style.height = height + 'px'
```
vojtechsimetka, line 20:
```suggestion
		textarea.style.height = `${height}px`
```

It is recommended to use templating rather than string operations.

vojtechsimetka (approved):

agazso (commented):
I tried this PR and there are a lot of improvements and the code looks good. However I just want to point out that it does not fix the first problem in #101, e.g. that the display name field can be still multiline.

sugarpusher:
@agazso 
I see what you're saying and I'm confused, I created the input-field.svelte component and replaced the display name field on the identity page. I tested on Brave, Chrome and Firefox and everything looks good locally. However when I tested just now from the vercel app, devtools shows a textarea field instead of the input field.  😮
I'll check what happened and update it.
Thanks for catching that!

sugarpusher:
@agazso 
I just tested again and it seems to be working fine, do you mind re-testing too?
https://waku-objects-playground-git-style-textarea-waku-objects.vercel.app/identity

agazso:
> @agazso I just tested again and it seems to be working fine, do you mind re-testing too? https://waku-objects-playground-git-style-textarea-waku-objects.vercel.app/identity

Yes, I still see this happening. Maybe it would be better to use a regular `input` with `type="text"` instead? From my experience `textarea` can be a bit difficult to work with.

vojtechsimetka:
Maybe the prblem is some caching somewhere?

agazso:
> Maybe the prblem is some caching somewhere?

We clarified in a call that we were testing different components :smile:

agazso (approved):
Works well!

PR #119, opened 2023-06-26
style: adjust spacings

sugarpusher:
- Add vertical spacing to Container and update components in use 
- Add label to input component and update components in use

### DIFF
--- a/src/lib/components/input-field.svelte
+++ b/src/lib/components/input-field.svelte
@@ -1,16 +1,16 @@
 <script lang="ts">
 	export let value = ''
-	export let placeholder = ''
+	export let placeholder: string | undefined = undefined
 	export let autofocus = false
 	export let disabled = false
 	export let pad = 0
-	export let label = ''
+	export let label: string | undefined = undefined
 
 	let inputField: HTMLInputElement
 </script>
 
 <div class="input-wrapper">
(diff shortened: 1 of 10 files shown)


src/lib/components/input-field.svelte
```diff
@@ -4,26 +4,44 @@
...
-	bind:value
-	bind:this={inputField}
-	on:keydown
-	on:keypress
-	on:keyup
-/>
+<div class="input-wrapper">
+	{#if label !== ''}
+		<span class="text-sm">{label}</span>
+	{/if}
```
vojtechsimetka, line 15: I think this should either be some named `slot` or the label should be potentially undefined `export let label: string | undefined = undefined`. I don't like having things set to empty strings and asserting to empty string.

vojtechsimetka (approved):
LGTM except for the default `label` value.

src/lib/components/input-field.svelte
sugarpusher: Cool, thank you. I'm not sure it should be a slot because it will never be anything other than a string that looks the exact same way, so having it as a prop protects the styling while being easier to declare in the code.
But I do agree regarding the empty string, I'll change that. 
Thanks!

PR #122, opened 2023-06-27
style: remove bubble when chat message is button

sugarpusher:
This PR attempts to solve the issue of bubbles being visible when the message is exclusively a button created by a Waku Object. It creates a ChatMessage component, keeping all styles in one place, and takes bool props for myMessage and bubble.


agazso (approved):
Tried it, works well!

PR #131, opened 2023-06-28
style: "new" Send Transaction screen

sugarpusher:


src/lib/objects/send_transaction/standalone.svelte
```diff
@@ -125,26 +127,37 @@
...
+				</Dropdown>
+			</div>
+			<p class="fiat text-sm">x EUR {amount ? 'now' : ''}</p>
+		</div>
+		<p
+			class={`balance ${
+				Number(amount) > Number(formatTokenAmount(token.amount, token.decimals)) ? 'text-bold' : ''
+			}`}
+		>
+			{#if Number(amount) > Number(formatTokenAmount(token.amount, token.decimals))}
```
vojtechsimetka, line 153: We should calculate this properly but doing this improperly is on me....

vojtechsimetka (approved):
Thanks, works great!

PR #135, opened 2023-06-29
fix: chat scroll on new message and on mount

### LINKED ISSUE #124: Chat does not scroll down when text is overflowing
In there is more text in the chat than what can fit on a screen I would expect that sending or receiving new text would automatically appear on the bottom, by automatically scrolling the view. Currently this is not the case.

vojtechsimetka:
Resolves #124

TODO:
- [x] scroll on new message
- [x] scroll on mount

### DIFF
--- a/src/routes/chat/[id]/+page.svelte
+++ b/src/routes/chat/[id]/+page.svelte
@@ -28,17 +28,8 @@
 	let autoscroll = true
 
 	beforeUpdate(() => {
-		if (browser && window) {
-			let body = document.body
-			let html = document.documentElement
-
-			let height = Math.max(
-				body.scrollHeight,
-				body.offsetHeight,
-				html.clientHeight,
-				html.scrollHeight,
-				html.offsetHeight,
-			)
+		if (browser && window && dev) {
+			const height = div.scrollHeight
 			autoscroll = height <= window.scrollY + window.innerHeight
 		}
 	})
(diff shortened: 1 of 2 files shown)


sugarpusher:
Scroll on new message and scroll on mount both look good to me, the only thing I'd probably add is scrolling when a message is sent but the view is not currently scrolled all the way down. So if you open the chat, scroll up a bit and send a message, it currently doesn't scroll down.

agazso:
Unfortunately for me it does not work correctly in Firefox :thinking:

vojtechsimetka:
@agazso it works for me on Firefox 114.0.2 (64-bit) on macOS . Maybe you have some older version?

agazso:
I was debugging this a bit and if I remove the `autoscroll` from the if condition check in the `afterUpdate` then it works correctly both on Firefox and Chromium. Then the `beforeUpdate` is not needed.

sugarpusher:
It works for me on firefox as well but I have the exact same setup as Vojtech

src/routes/chat/[id]/+page.svelte
```diff
@@ -28,15 +28,40 @@
...
+			let body = document.body
+			let html = document.documentElement
+
+			let height = Math.max(
+				body.scrollHeight,
+				body.offsetHeight,
+				html.clientHeight,
+				html.scrollHeight,
+				html.offsetHeight,
+			)
```
Comment (agazso) [inline src/routes/chat/[id]/+page.svelte:41]:
```suggestion
		if (browser && window && dev) {
			const height = div.scrollHeight
```

This works for me on all browsers and to me the code is more understandable and as far as I understand the previous version it does the same thing.

vojtechsimetka:
This is a bit jumpy on safari as I type. I'll merge as is and we can tackle it in a separate PR.

PR #139, opened 2023-06-29
feat: add transaction support for waku adapter

agazso:
This adapts the changes from the Firebase adapter to the Waku adapter. The current implementation is still considered a simplification, because it does not use the blockchain for transactions, it only simulates it. Therefore I did not try to achieve completeness, because most of the code will change when the real transactions are implemented.

Known issues:
- There is a fundemantal difference between the Firebase and the Waku adapter: with the Firebase everything is written in the database and there are subscriptions on the database that updates the local stores. In the Waku adapter this is the other way around. Because of this currently the balances are not updated automatically. This is however subject to change when real blockchain transactions are implemented.

### DIFF
--- a/src/lib/adapters/waku/index.ts
+++ b/src/lib/adapters/waku/index.ts
@@ -19,7 +19,8 @@ import { ipfs, IPFS_GATEWAY } from '../firebase/connections'
 import { get } from 'svelte/store'
 import { objectStore, type ObjectState, objectKey } from '$lib/stores/objects'
 import { lookup } from '$lib/objects/lookup'
-import type { Token } from '$lib/stores/balances'
+import { balanceStore, type BalanceState, type Token } from '$lib/stores/balances'
+import { TokenDbSchema, type TokenDb } from '../firebase/schemas'
 
 function createChat(chatId: string, user: User, address: string): string {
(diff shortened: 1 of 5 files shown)


src/lib/adapters/waku/index.ts
```diff
@@ -118,6 +119,33 @@ async function storeObjectStore(waku: LightNode, address: string, objectStore: O
...
+
+async function storeBalances(waku: LightNode, address: string, balanceState: BalanceState) {
+	return await storeDocument(
+		waku,
+		'balances',
+		address,
+		balanceState.balances.map((token) => ({ ...token, amount: token.amount.toString() })),
+	)
+}
+
```
vojtechsimetka, line 148: Do you want to store balances in waku? Imho this is quite temporary and it may just complicate our life. I would either skip this for now, or implement at least reading the balances from the chain (we don't need to necessarily do the transaction sending for now).

agazso: I agree that we should not store the balances in waku. The only reason I did this so that the balances actually change when there is a transaction. We can skip this for now.

vojtechsimetka (approved):

PR #146, opened 2023-07-03
fix: textarea resize on content change

### LINKED ISSUE #130: Chat screen textarea doesn't resize upon send
When a multiline message is sent by clicking the send icon, the message is sent correctly but the textarea field does not go back to its original size until it's typed in again. This does not happen when sending the message by hitting enter.

vojtechsimetka:
resolves #130


agazso (approved):
Works well.

When testing this I found another problem: adding a newline to the input field does not result in a newline in the message. Maybe the best is to open a new issue about that.

PR #150, opened 2023-07-04
chore: update deps

vojtechsimetka:
I am primarily interested in updating svelte to v4 but it seems we can update everything. The reason for updating svelte is dev performance and especially false code issues reporting which happen frequently in v3.


agazso (approved):
I tested this and it seems to be working correctly. It reported a lot of A11y issues when building though, not sure what to do with that:

```
10:39:24 AM [vite-plugin-svelte] /src/lib/components/select.svelte:10:0 A11y: <div> with click handler must have an ARIA role
 8: 
 9: <!-- svelte-ignore a11y-click-events-have-key-events -->
10: <div class={`root ${cls} ${disabled ? 'disabled' : ''}`} on:click>
    ^
11:   <div class="label">
12:     {label}

```

vojtechsimetka:
Yeah ideally we should actually resolve these. They are good pointers what to add in terms of accessibility.

PR #153, opened 2023-07-05
chore: disable adding images on chat

sugarpusher:


src/lib/components/dropdown-item.svelte
```diff
@@ -37,7 +37,7 @@
 		}
 
 		&.disabled {
-			color: var(--grey-200);
+			color: var(--gray20);
```
vojtechsimetka: This seems to be typo. I also found some other instances of `gray` and not `grey`

vojtechsimetka: ah it's US English. You learn something new every day!

sugarpusher: yup I think there's some leftover from initial copying from Kurate that I find every once in a while. I'll do a project wide search. But I also just changed that one to something else 😊

vojtechsimetka (approved):

PR #155, opened 2023-07-05
chore: remove Request Transaction object from list and rename Send Transaction to Payggy

sugarpusher:


agazso (approved):
Wow, this change made it look like a real app!

One thing I noticed that in the waku objects list it still shows a random icon for Payggy. Now that we have the right icon we could show that.

agazso:
> Wow, this change made it look like a real app!
> 
> One thing I noticed that in the waku objects list it still shows a random icon for Payggy. Now that we have the right icon we could show that.

Well, it turns out that it was not only this PR that made a big change, but combined with the rest I did not see before. Still, it looks great!

sugarpusher:
> 

Yeah @vojtechsimetka is on this 😊

Issue #132, opened 2023-06-29
Payggy: Calculate and check if user has enough money properly


vojtechsimetka:
_Originally posted by @vojtechsimetka in https://github.com/logos-innovation-lab/waku-objects-playground/pull/131#discussion_r1246148093_

Issue #127, opened 2023-06-28
Payggy: implement transaction status checking


vojtechsimetka:
For now the WO only has info about the tx hash but not whether the transaction succeeded

PR #156, opened 2023-07-05
feat: blockchain transactions

agazso:
MVP for blockchain transactions. Automatically updates the balance when a transaction was sent/received.

It is already quite big, so it would be good to merge. There are known issues, planned to be fixed in later PRs:

- [x] only Firebase, no Waku yet (it is be easy to add)
- [ ] it only works with a locally running RPC node (e.g. hardhat), but I tested it successfully on Sepolia testnet (https://github.com/logos-innovation-lab/waku-objects-playground/issues/170)
- [ ] it only works with the native token (https://github.com/logos-innovation-lab/waku-objects-playground/issues/168)
- [ ] there is no transaction fee management (https://github.com/logos-innovation-lab/waku-objects-playground/issues/132)
- [ ] there is no pending state or error handling implemented for the transaction (https://github.com/logos-innovation-lab/waku-objects-playground/issues/127)

Other observations
- [ ] the `onMessage` implementation relies on the global `adapter`, it would be better if that (or a subset) would be passed in as argument
- [ ] it would be better if the messages could be discriminated with a `type` field rather than trying to parse them
- [ ] the `Adapter` type relies on a `HDNodeWallet` typed wallet to be passed in almost all function, but it is not used, only the `address` field from it. It seems it would be better to just use the `address` instead.
- [ ] it would be good to come up with some strategy when dealing with Tokens, currently there are three slightly different versions (in the balance store, in Firebase and in the transaction message). Also it would be good to define how new tokens can be added by the user.
- [x] the current precision of 2 in `formatTokenAmount` seems to be a bit limiting when testing, also this may be problematic with mainnet ETH :smile: (#169)

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -6,6 +6,7 @@ const sveltePreprocessor = sveltePreprocess()
 
 const config: KnipConfig = {
 	ignore: ['**/*.d.ts'],
+	ignoreDependencies: ['hardhat'],
 	paths: {
 		// This ain't pretty, but Svelte basically does the same
 		'$app/*': ['node_modules/@sveltejs/kit/src/runtime/app/*'],
--- a/knip.ts
+++ b/knip.ts
@@ -5,7 +5,7 @@ import type { KnipConfig } from 'knip'
 const sveltePreprocessor = sveltePreprocess()
 
 const config: KnipConfig = {
-	ignore: ['**/*.d.ts'],
+	ignore: ['**/*.d.ts', '**/schemas.ts', 'src/lib/objects/hello-world/index.ts'],
 	ignoreDependencies: ['hardhat'],
 	paths: {
 		// This ain't pretty, but Svelte basically does the same
--- a/knip.ts
+++ b/knip.ts
@@ -6,7 +6,6 @@ const sveltePreprocessor = sveltePreprocess()
 
 const config: KnipConfig = {
 	ignore: ['**/*.d.ts', '**/schemas.ts', 'src/lib/objects/hello-world/index.ts'],
-	ignoreDependencies: ['hardhat'],
 	paths: {
 		// This ain't pretty, but Svelte basically does the same
 		'$app/*': ['node_modules/@sveltejs/kit/src/runtime/app/*'],
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
 		"test:unit": "vitest",
 		"lint": "prettier --plugin-search-dir . --check . && eslint .",
 		"format": "prettier --plugin-search-dir . --write . && eslint . --fix",
-		"knip": "knip"
+		"knip": "knip",
+		"start:blockchain": "hardhat node"
 	},
 	"devDependencies": {
 		"@bonosoft/sveltekit-qrcode": "^0.0.3",
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
 		"eslint-plugin-svelte3": "^4.0.0",
 		"ethers": "^6.6.2",
 		"firebase": "^9.23.0",
+		"hardhat": "^2.16.1",
 		"ipfs-http-client": "^60.0.0",
 		"knip": "^2.14.3",
 		"prettier": "^2.8.8",
--- a/src/lib/adapters/firebase/index.ts
+++ b/src/lib/adapters/firebase/index.ts
@@ -131,7 +131,6 @@ export default class FirebaseAdapter implements Adapter {
 
 						// update the object store if there is an incoming data message
 						parseRes.data.messages.forEach((message) => {
-							console.debug('onSnapshot', { message })
 							if (message.type === 'data') {
 								const descriptor = lookup(message.objectId)
 								const key = objectKey(message.objectId, message.instanceId)
--- a/src/lib/adapters/firebase/index.ts
+++ b/src/lib/adapters/firebase/index.ts
@@ -40,6 +40,7 @@ import { objectKey, objectStore } from '$lib/stores/objects'
 import { lookup } from '$lib/objects/lookup'
 import { type Unsubscriber, get } from 'svelte/store'
 import { sleep } from '../utils'
+import { sendTransaction } from '../transaction'
 
 export default class FirebaseAdapter implements Adapter {
 	protected userSubscriptions: Array<() => unknown> = []
@@ -301,6 +302,7 @@ export default class FirebaseAdapter implements Adapter {
 			decimals: 18,
 			amount: 7843900000000000000000n.toString(),
 			image: 'https://s2.coinmarketcap.com/static/img/coins/64x64/4943.png',
+			address: '0x0000000000000000000000000000000000000000',
 		}
 
 		setDoc(daiDoc, daiData, { merge: true })
@@ -334,23 +336,7 @@ export default class FirebaseAdapter implements Adapter {
 
 		if (!address) throw new Error('Address is missing')
 
-		const tx = {
-			from: address,
-			to,
-			token: {
-				amount: token.amount.toString(),
-				name: token.name,
-				symbol: token.symbol,
-				decimals: token.decimals,
-			},
-			timestamp: Date.now(),
-			fee: {
-				amount: fee.amount.toString(),
-				symbol: fee.symbol,
-				name: fee.name,
-				decimals: fee.decimals,
-			},
-		}
+		/*
 
 		// Update balances
 		const balanceFromDoc = doc(db, `users/${address}/balances/${token.symbol.toLowerCase()}`)
@@ -373,11 +359,32 @@ export default class FirebaseAdapter implements Adapter {
 		// Calculate and deduct fee
 		const feeToken = (await getDoc(feeDoc)).data()
 		if (!feeToken) throw new Error('Fee not found')
-		setDoc(feeDoc, { amount: (BigInt(feeToken.amount) - fee.amount).toString() }, { merge: true })
+		setDoc(feeDoc, { amount: (BigInt(feeToken.amount) - fee.amount).toString() }, { merge: true })	
+		*/
+
+		const tx = await sendTransaction(wallet, to, token.amount, fee.amount)
 
 		const txCollection = collection(db, `transactions`)
-		const res = await addDoc(txCollection, tx)
-		return res.id
+		const txData = {
+			from: address,
+			to,
+			token: {
+				amount: token.amount.toString(),
+				name: token.name,
+				symbol: token.symbol,
+				decimals: token.decimals,
+			},
+			timestamp: Date.now(),
+			fee: {
+				amount: fee.amount.toString(),
+				symbol: fee.symbol,
+				name: fee.name,
+				decimals: fee.decimals,
+			},
+		}
+
+		const res = await addDoc(txCollection, txData)
+		return tx.hash
 	}
 
 	estimateTransaction(): Promise<Token> {
--- a/src/lib/adapters/firebase/index.ts
+++ b/src/lib/adapters/firebase/index.ts
@@ -182,6 +182,7 @@ export default class FirebaseAdapter implements Adapter {
 		const subscribeBalances = onSnapshot(balanceCollection, (res) => {
 			const balances: Token[] = []
 			res.docs.forEach((d) => {
+				console.debug({ d, data: d.data() })
 				const parseRes = TokenDbSchema.safeParse(d.data())
 				if (parseRes.success) {
 					const token: Token = parseRes.data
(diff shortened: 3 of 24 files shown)


src/lib/adapters/firebase/index.ts
```diff
@@ -373,11 +379,34 @@ export default class FirebaseAdapter implements Adapter {
...
+			timestamp: Date.now(),
+			fee: {
+				amount: fee.amount.toString(),
+				symbol: fee.symbol,
+				name: fee.name,
+				decimals: fee.decimals,
+			},
+		}
+
+		const res = await addDoc(txCollection, txData)
```
filoozom, line 408: Why are we storing this? And why is `res` assigned but not used?

src/lib/adapters/firebase/index.ts
```diff
@@ -301,11 +305,28 @@ export default class FirebaseAdapter implements Adapter {
...
+		}
+
+		const nativeTokenAmount = await getBalance(address)
+
+		const ethDoc = doc(db, `users/${address}/balances/${token.symbol.toLowerCase()}`)
+		const ethData: { amount: string } = {
+			amount: nativeTokenAmount.toString(),
+		}
+
+		setDoc(ethDoc, ethData, { merge: true })
```
filoozom, line 327: Why are we storing this?

vojtechsimetka:
> * [ ]  the current precision of 2 in `formatTokenAmount` seems to be a bit limiting when testing, also this may be problematic with mainnet ETH 😄

Probably makes sense to define a function on the `Token` interface that is `.toString()` and formats it to x significant digits.

package.json
```diff
@@ -30,6 +30,7 @@
 		"eslint-plugin-svelte3": "^4.0.0",
 		"ethers": "^6.6.2",
 		"firebase": "^9.23.0",
+		"hardhat": "^2.16.1",
```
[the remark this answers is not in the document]
agazso: I tried npx hardhat and it seems that it only works if it is installed. We can use something else, this just worked for me at first without being familiar with it. 

```bash
% npx hardhat node
888    888                      888 888               888
888    888                      888 888               888
888    888                      888 888               888
8888888888  8888b.  888d888 .d88888 88888b.   8888b.  888888
888    888     "88b 888P"  d88" 888 888 "88b     "88b 888
888    888 .d888888 888    888  888 888  888 .d888888 888
888    888 888  888 888    Y88b 888 888  888 888  888 Y88b.
888    888 "Y888888 888     "Y88888 888  888 "Y888888  "Y888

Welcome to Hardhat v2.16.1

✔ What do you want to do? · Create an empty hardhat.config.js
Config file created

You need to install hardhat locally to use it. Please run:

pnpm add -D "hardhat@^2.16.1"


Give Hardhat a star on Github if you're enjoying it!

     https://github.com/NomicFoundation/hardhat
Start time 21:34:55, end time 21:35:15, elapsed 20s

% npx hardhat node
Error HH12: Trying to use a non-local installation of Hardhat, which is not supported.
Please install Hardhat locally using npm or Yarn, and try again.

For more info go to https://hardhat.org/HH12 or run Hardhat with --show-stack-traces
```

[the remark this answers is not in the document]
vojtechsimetka: Alternatively, you can fix it by adding script to the `package.json` which knip will detect. Something like `"start:blockchain": "hardhat"`

src/routes/identity/account/+page.svelte
```diff
@@ -27,6 +28,28 @@
...
+		const address = $walletStore.wallet?.address
+		if (address === undefined) return
+
+		const balance = await getBalance(address)
+		console.debug({ balance, $balanceStore })
+
+		balanceStore.update((balanceState) => ({
+			...balanceState,
+			balances: balanceState.balances.map((value) => {
+				if (value.address) {
```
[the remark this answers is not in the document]
agazso: Currently the `getBalance` only returns the balance of the native token. The `value.address` is the contract address of the token and it is `undefined` when it's the native token. Therefore this code just fetches the native token balance and updates it, while keeping the other tokens unchanged.

In order to make the non-native tokens to work, there needs to be added a function that fetches the token balance from the contract using the address. This is currently missing and this could be done as a separate issue. I will open one.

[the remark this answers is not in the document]
agazso: https://github.com/logos-innovation-lab/waku-objects-playground/issues/168

package.json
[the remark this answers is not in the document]
agazso: > Alternatively, you can fix it by adding script to the `package.json` which knip will detect. Something like `"start:blockchain": "hardhat"`

That did not work for me, I still had to ignore hardhat to avoid errors.

src/lib/adapters/firebase/index.ts
agazso: I opened an issue about this topic: https://github.com/logos-innovation-lab/waku-objects-playground/issues/171

src/lib/adapters/firebase/index.ts
agazso: I opened an issue about this topic: https://github.com/logos-innovation-lab/waku-objects-playground/issues/171

agazso:
> > * [ ]   the current precision of 2 in `formatTokenAmount` seems to be a bit limiting when testing, also this may be problematic with mainnet ETH smile
> 
> Probably makes sense to define a function on the `Token` interface that is `.toString()` and formats it to x significant digits.

I opened an issue about this: https://github.com/logos-innovation-lab/waku-objects-playground/issues/169

src/lib/adapters/waku/index.ts
```diff
@@ -453,4 +458,22 @@ export default class WakuAdapter implements Adapter {
 		}
 		balanceStore.set(balancesState)
 	}
```
vojtechsimetka, line 460: Do we still need this?

src/lib/objects/payggy/chat.svelte
```diff
@@ -36,8 +25,7 @@
 			console.error(res.error)
 		}
 	}
-	$: myMessage = data?.from === args.profile.address
-	// FIXME: will not work for group chats
+	$: myMessage = message.fromAddress === args.profile.address
 	$: otherUser = args.users.find((u) => u.address !== args.profile.address)
```
vojtechsimetka, line 29:
```suggestion
	$: myMessage = message.fromAddress === args.profile.address
	// FIXME: will not work for group chats
	$: otherUser = args.users.find((u) => u.address !== args.profile.address)
```
I'd prefer to keep the note

src/lib/objects/payggy/index.ts
```diff
@@ -15,7 +16,19 @@ export const payggyDescriptor: WakuObjectDescriptor = {
...
 
-	onMessage: (address, store, message) => {
+	onMessage: (address, adapter, store, message) => {
+		if (message?.data) {
+			const res = MessageDataSendSchema.safeParse(message.data)
+			if (res.success) {
+				const token = {
+					...res.data.token,
+					name: res.data.token.symbol,
+					amount: BigInt(0), // the amount is not really necessary for checkBalance
```
vojtechsimetka, line 26: maybe we can have a `Token` type that does not contain `amount` field?

src/lib/objects/payggy/standalone.svelte
```diff
@@ -18,43 +18,53 @@
...
+	}
+	let toUser =
+		args.users.find((user) => user.address !== args.profile.address) || throwError('invalid user')
+	$: if (amount && token) {
 		const tokenToTransfer = { ...token, amount: BigInt(Number(amount) * 10 ** token.decimals) }
-		args.estimateTransaction(store.toUser.address, tokenToTransfer).then((f) => (fee = f))
+		args.estimateTransaction(toUser.address, tokenToTransfer).then((f) => (fee = f))
 	}
-	$: if (!amount && store.view === 'overview') history.back()
+	// $: if (!amount && store && store.view === 'overview') history.back()
```
vojtechsimetka, line 57: This is here in case user refreshes which makes the `amount` an empty string. The assumption was, that user should be navigated to the page, where they fill the amount in.

src/lib/objects/ui.svelte
```diff
@@ -13,56 +18,60 @@
...
+	let store: unknown
+	$: store = $objectStore.objects.get(objectKey(objectId, instanceId))
+	let tokens: Token[]
 	$: tokens = $balanceStore.balances
-	$: nativeToken = tokens.find((token) => token.symbol === 'ETH')
-	$: chat = $chats.chats.get($page.params.id)
-	$: fromUser = chat?.users.find((user) => user.address === $walletStore.wallet?.address)
-	$: toUser = chat?.users.find((user) => user.address !== $walletStore.wallet?.address)
+	let chat: Chat
+	$: chat = $chats.chats.get(chatId) || throwError('chat not found')
```
vojtechsimetka, line 33: This likely means that on page refresh, it throws error and fails to render the page.

src/routes/identity/account/+page.svelte
```diff
@@ -63,6 +85,8 @@
 	<Container align="center" direction="row" gap={6} justify="center" padX={24}>
 		Assets <Badge dark>{$balanceStore.balances.length}</Badge>
 	</Container>
+	<Button on:click={getBalances}>Fetch balances</Button>
```
vojtechsimetka, line 88: Can't we fetch them `onMount` in the right page? Or on the whole page load. I don't like this here.

vojtechsimetka (commented):
Great progress, thanks!

knip.ts
```diff
@@ -5,7 +5,8 @@ import type { KnipConfig } from 'knip'
 const sveltePreprocessor = sveltePreprocess()
 
 const config: KnipConfig = {
-	ignore: ['**/*.d.ts'],
+	ignore: ['**/*.d.ts', '**/schemas.ts', 'src/lib/objects/hello-world/index.ts'],
```
vojtechsimetka, line 8: Why are the `'**/schemas.ts', 'src/lib/objects/hello-world/index.ts'` ignored?

src/lib/adapters/waku/index.ts
agazso: Yeah, because this is at the intersection of two issues:
https://github.com/logos-innovation-lab/waku-objects-playground/issues/168 and https://github.com/logos-innovation-lab/waku-objects-playground/issues/171

First of all we need to initialize the balances somewhere when an account is created. Currently there is no real support for multiple currencies (at least not with the transactions) but the UI supports them and removing this code would only show the native token.

Secondly, with the Waku adapter the balance is not stored on the network, so it has to come from somewhere. In the case of the native token it comes from the blockchain, and for other currencies it is hardcoded. I think we should first fix the support for non-native tokens and then the Firebase should follow the same pattern. Then we can get rid of this function.

src/lib/objects/payggy/index.ts
agazso: Good idea, I opened an issue about this: #177 

I would not include that in this PR.

knip.ts
agazso: Because knip reports unused exports when only a type is imported, which is the case with the schema files and the `HelloWorldStore` in case of the hello world `index.ts`.

src/lib/adapters/waku/index.ts
vojtechsimetka: I see. But for the waku adapter we can for now only support native token. That's completely fine, we don't need to have additional token. But I am also OK with keeping this code here.

src/lib/objects/payggy/standalone.svelte
vojtechsimetka: I can't replicate that issue. But I did notice that the URL change is wrong now on main. It goes from `/chat/koKPzjnB57rInHObAJMY/object/payggy/new` to `/chat/koKPzjnB57rInHObAJMY/object/new/new/overview`

knip.ts
vojtechsimetka: I see. This is a good one to report to the `knip` dev. I'll do that

src/routes/identity/account/+page.svelte
```diff
@@ -29,26 +30,12 @@
...
-					return {
-						...value,
-						amount: balance,
-					}
-				}
-			}),
-		}))
-	}
+		adapter.initializeBalances(address)
+	})
```
vojtechsimetka: In the `firebase` adapter I do this when the user logs in and the balances are fetched. I wonder if that would also work. This way, we don't need to go to `/identity/account` to initialize the balances. It happens after creating/restoring wallet.

https://github.com/logos-innovation-lab/waku-objects-playground/blob/main/src/lib/adapters/firebase/index.ts#L193

vojtechsimetka (approved):

src/routes/identity/account/+page.svelte
vojtechsimetka: Oh I see this is also resolved no and is in `onLogIn`

agazso: It's still there and also in the `waku` adapter. However if the balance is only checked once the user logs in then the account page may show an outdated balance when you see it. I had this exact problem during testing and that's why I added the button there, but the `onMount` version is better.

src/lib/adapters/waku/index.ts
agazso: I prefer to keep the feature parity, because there are less surprises later.

src/lib/objects/payggy/standalone.svelte
agazso: I added a comment to the issue #178 explaining the problem. As I write there it would be better to come up with a generic solution to this, because it is very easy to introduce a change that currently breaks the page when reloading.

I also opened a new issue for the URL change problem: #179

Issue #125, opened 2023-06-28
Payggy: Implement the missing payggy screens


vojtechsimetka:
https://www.sketch.com/s/d25bf557-2cc4-4480-9343-cf5255b32a20

PR #160, opened 2023-07-06
feat: details view on waku object

vojtechsimetka:
Part of #125

View in chat with icon that can be expanded
Screenshot 2023-08-04 at 18 01 07

Detail view
Screenshot 2023-08-04 at 18 01 01

### DIFF
--- a/src/lib/adapters/transaction.ts
+++ b/src/lib/adapters/transaction.ts
@@ -128,3 +128,9 @@ export async function getTransactionReceipt(txHash: string): Promise<Transaction
 	const receipt = await provider.getTransactionReceipt(txHash)
 	return receipt
 }
+
+export async function getTransactionTimestamp(blocknum: number): Promise<number> {
+	const provider = getProvider()
+	const block = await provider.getBlock(blocknum)
+	return block?.timestamp ?? Date.now()
+}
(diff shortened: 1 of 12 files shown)


sugarpusher:
I also can't click the details button from the recipient's bubble

agazso:
I tested and found some issues:
- [ ] the details page layout seems to be broken (also on your picture, but I attached one to clarify what I refer to)
- [ ] reloading the details page results in 500 Internal error
- [ ] I think by design the gray area on the widget is clickable, not just the icon. Otherwise the icon is too small, so the click area should be larger (for phones)
- [ ] I get plenty of Zod errors when I close the details page (see screenshot)

For me clicking the recipient details button worked.

Screenshot from 2023-08-04 18-26-55

Screenshot from 2023-08-04 18-32-21

agazso:
And to clarify, I would only investigate the Zod issue, the others are non-blockers. It would be good to open issues for them though.

vojtechsimetka:
The Zod issues are there because you have some existing paygy transactions and the paygy store has changed (it now has 2 extra fields) Thats why it complains and the old paygy fail to render.

agazso (approved):

Issue #167, opened 2023-07-09
Unnecessary scrollbar on Backup Recovery Phrase screen


agazso:
I see a scrollbar that seems unnecessary on the Backup Recovery Phrase screen. Only reproducible in Chrome, but not in Firefox (both on Linux). Maybe it is related to the font or the font size, because to me the browsers are using different fonts.

Update: on Firefox the text fits in two lines, in Chrome it's broken in three, that's why the scrollbar appears. I think it would be better if the box could expand a bit, there is space for it.

![Screenshot from 2023-07-09 20-01-27](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/1182dbba-7041-411a-9af0-7d608c02a198)

PR #173, opened 2023-07-10
chore: replace textarea with ReadonlyText

sugarpusher:
Hopefully solves #167

I don't have Linux so I can't reproduce the reported bug, but the field shouldn't be a `<textarea>` anyway so I replaced it with the ReadonlyText component. Hopefully that solves it.


agazso (approved):
Works well!

vojtechsimetka (approved):

PR #174, opened 2023-07-10
feat: to significant digits token amount formatting

### LINKED ISSUE #169: Precision of 2 in `formatTokenAmount` is limiting when testing
> * [ ]  the current precision of 2 in `formatTokenAmount` seems to be a bit limiting when testing, also this may be problematic with mainnet ETH 😄

Probably makes sense to define a function on the `Token` interface that is `.toString()` and formats it to x significant digits.

_Originally posted by @vojtechsimetka in https://github.com/logos-innovation-lab/waku-objects-playground/issues/156#issuecomment-1623279723_

vojtechsimetka:
Resolves #169 

This add token manipulation methods which have full bigint precision

Method `toBigInt` - validates and parses string representing token into bigint representation
     ✓ with value=1.234567890123456789 and decimals=18 should return 1234567890123456789
     ✓ with value=1.234567890123456789 and decimals=6 should return 1234567
     ✓ with value=0.00012345 and decimals=3 should return 0
     ✓ with value=0.0009999 and decimals=3 should return 0
     ✓ with value=0.001 and decimals=3 should return 1
     ✓ with value=0 and decimals=3 should return 0
     ✓ with value=0.0 and decimals=3 should return 0
     ✓ with value=1 and decimals=3 should return 1000
     
Method `toSignificant` - formats token representation with amount as bigint and decimals to string which has some minimal fractional significant digits
     ✓ with value=1.234567890123456789 and minDecSigDigits=0 should return 1
     ✓ with value=1.234567890123456789 and minDecSigDigits=4 should return 1.2345
     ✓ with value=0.001234567890123456 and minDecSigDigits=4 should return 0.001234
     ✓ with value=1 and minDecSigDigits=2 should return 1.00
     ✓ with value=1.234 and minDecSigDigits=undefined should return 1.23
     ✓ with value=0.00001234 and minDecSigDigits=undefined should return 0.000012
     ✓ with value=0.0000001000 and minDecSigDigits=3 should return 0.0000001

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,7 +1,0 @@
-import { describe, it, expect } from 'vitest'
-
-describe('sum test', () => {
-	it('adds 1 + 2 to equal 3', () => {
-		expect(1 + 2).toBe(3)
-	})
-})
--- a/src/lib/components/asset.svelte
+++ b/src/lib/components/asset.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import Container from '$lib/components/container.svelte'
-	import { formatTokenAmount } from '$lib/utils/format'
+	import { toSignificant } from '$lib/utils/format'
 	export let name: string
 	export let token: string
 	export let amount: bigint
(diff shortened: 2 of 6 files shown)


sugarpusher (approved):
LGTM

agazso (approved):

PR #180, opened 2023-07-11
style: fix corners and spacing on chat bubbles

sugarpusher:
Known issue: This won't work on Firefox. 

CSS doesn't have a "previous" sibling selector, so I had to use the :has() relational pseudo-class to achieve the same result. Unfortunately, :has() is not yet supported on Firefox (see [here](https://caniuse.com/?search=%3Ahas)). 

The alternative solution would be to use javascript to achieve the desired result, however I believe that since this is a visual enhancement, the performance cost might not be worth it.

### DIFF
--- a/src/lib/components/chat-message.svelte
+++ b/src/lib/components/chat-message.svelte
@@ -4,11 +4,7 @@
 
 	const isFF = () => {
 		let browserInfo = navigator.userAgent
-		if (browserInfo.includes('Firefox')) {
-			return true
-		} else {
-			return false
-		}
+		return browserInfo.includes('Firefox'))
 	}
 </script>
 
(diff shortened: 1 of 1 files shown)


src/lib/components/chat-message.svelte
```diff
@@ -1,9 +1,18 @@
...
 	export let bubble = false
 	export let myMessage = false
+
+	const isFF = () => {
+		let browserInfo = navigator.userAgent
+		if (browserInfo.includes('Firefox')) {
+			return true
+		} else {
+			return false
+		}
```
agazso, line 11:
```suggestion
		return browserInfo.includes('Firefox'))
```

I would simplify it to this, because you can infer the type and value from the function name.

agazso (approved):

src/lib/components/chat-message.svelte
sugarpusher: oh yeah that's much smarter 😝
Thank you!!

Issue #177, opened 2023-07-11
Refactor `Token` type so that it does not contain `amount` field


agazso:
maybe we can have a `Token` type that does not contain `amount` field?

_Originally posted by @vojtechsimetka in https://github.com/logos-innovation-lab/waku-objects-playground/pull/156#discussion_r1257545267_

PR #181, opened 2023-07-12
feat: transaction status

agazso:
With this PR the Payggy object is able to set the status of the transactions on the backend correctly and update it with messages. It only sends the transaction hash in the messages and the peer in the chat reads the transaction from the blockchain based on that.

What is done:
- [x] moved balance out of adapter
- [x] made the Waku and Firebase adapter iterators async to enforce in-order evaluation
    - this was done in order to avoid possible race conditions 
- [x] made the `onMessage` callback async, no longer returns the updated store
    - instead there is a new `updateStore` function passed as argument
- [x] added Sepolia and Chiado testnet configs
  - [x] made Chiado the default 
- [x] made some minor visual fixes in the assets and the dev page
- [x] added transaction states and the Payggy app correctly displays them (except the error, design is missing for that)

Known issues:
- [ ] Payggy still relies on the messages to arrive for updating the transaction state. If the messages does not arrive then the transaction state won't change. It would be good to add some kind of mechanism to keep track of the active transactions locally once they arrived so that the app does not have to rely on messages.
- [ ] I haven't fixed the Token issue (#177) in this PR, but it was very annoying. I suggest removing the `amount` from the `Token` type, and creating an `Amount` type which could have a `token` and a `value` property.
- [ ] Chiado RPC endpoints are currently not working. It seems that Nethermind has an issue with `ethers@v6`. `ethers@v5` would work. Workaround was to set up an alternative endpoint running a different execution client. (https://github.com/NethermindEth/nethermind/issues/5921 and https://github.com/ethers-io/ethers.js/issues/3835)
- [ ] Missing error handling all over the place, especially in the design for the Payggy chat widget.
  - One example: https://github.com/logos-innovation-lab/waku-objects-playground/issues/182
- [ ] Currently there are adapter-like functionalities all over the code (there are Firebase & Waku, balance, transaction and the object adapter). It would be good to consolidate these at some point.
- [ ] There are several `onMessage` arguments now. It would be better to just have two arguments, one `DataMessage` and one `WakuObjectArgs`. I would also rename the latter to `WakuObjectContext`.
- [ ] During testing it happened to me that the two chat apps were on different blockchain network, but they were connected on the same messaging network. Obviously the transactions did not work. Maybe we could include the chain ID in the messages to avoid similar situations.

### DIFF
--- a/src/lib/adapters/firebase/index.ts
+++ b/src/lib/adapters/firebase/index.ts
@@ -20,14 +20,7 @@ import {
 import { profile } from '$lib/stores/profile'
 import { contacts, type User } from '$lib/stores/users'
 
-import {
-	ChatDbSchema,
-	TokenDbSchema,
-	UserDbSchema,
-	type ProfileDb,
-	type TokenDb,
-	ObjectDbSchema,
-} from './schemas'
+import { ChatDbSchema, UserDbSchema, type ProfileDb, ObjectDbSchema } from './schemas'
 
 import { formatAddress } from '$lib/utils/format'
 import { db, ipfs, IPFS_GATEWAY } from './connections'
@@ -39,8 +32,9 @@ import { objectKey, objectStore } from '$lib/stores/objects'
 import { lookup } from '$lib/objects/lookup'
 import { type Unsubscriber, get } from 'svelte/store'
 import { sleep } from '../utils'
-import { defaultBlockchainNetwork, getBalance, sendTransaction } from '$lib/adapters/transaction'
+import { sendTransaction } from '$lib/adapters/transaction'
 import { makeWakuObjectAdapter } from '$lib/objects/adapter'
+import { initializeBalances } from '$lib/adapters/balance'
 
 export default class FirebaseAdapter implements Adapter {
 	protected userSubscriptions: Array<() => unknown> = []
@@ -188,23 +182,7 @@ export default class FirebaseAdapter implements Adapter {
 		})
 		this.userSubscriptions.push(subscribeUsers)
 
-		const balanceCollection = collection(db, `users/${address}/balances`)
-		const subscribeBalances = onSnapshot(balanceCollection, (res) => {
-			const balances: Token[] = []
-			res.docs.forEach((d) => {
-				const parseRes = TokenDbSchema.safeParse(d.data())
-				if (parseRes.success) {
-					const token: Token = parseRes.data
-					balances.push(token)
-				} else {
-					console.error(parseRes.error.issues)
-				}
-			})
-			// TODO: Handle this better, every new signed in user should have some balances set
-			if (balances.length === 0) this.initializeBalances(address)
-			balanceStore.set({ balances, loading: false })
-		})
-		this.userSubscriptions.push(subscribeBalances)
+		initializeBalances(address)
 	}
 
 	onLogOut() {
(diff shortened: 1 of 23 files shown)


src/lib/objects/schemas.ts
```diff
@@ -17,3 +17,27 @@ export const UserSchema = z.object({
...
+		symbol: z.string(),
+		decimals: z.number().int().positive(),
+	}),
+	to: AddressSchema,
+	from: AddressSchema,
+	fee: z.object({
+		amount: z.string(),
+		symbol: z.string(),
+		decimals: z.number().int().positive(),
+	}),
```
vojtechsimetka, line 33: Let's keep it as is right now, but later on we should think about whether we don't want to use `token address` instead of `symbol` as symbols can be ambiguous (although they should not)

vojtechsimetka (commented):
This PR changes the native balance to xDAI for me but the payggy still show ETH as native currency (not sure why) so I can not send funds.

src/lib/objects/payggy/standalone.svelte
```diff
@@ -59,26 +60,22 @@
...
 			})
+
 			history.go(-3)
+
+			setTimeout(async () => {
+				await args.waitForTransaction(transactionHash)
+				await args.send({
+					hash: transactionHash,
+				})
+			}, 0)
```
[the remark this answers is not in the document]
agazso: That's a good point.

src/lib/adapters/firebase/index.ts
```diff
@@ -184,23 +182,7 @@ export default class FirebaseAdapter implements Adapter {
...
-				} else {
-					console.error(parseRes.error.issues)
-				}
-			})
-			// TODO: Handle this better, every new signed in user should have some balances set
-			if (balances.length === 0) this.initializeBalances(address)
-			balanceStore.set({ balances, loading: false })
-		})
-		this.userSubscriptions.push(subscribeBalances)
+		initializeBalances(address)
```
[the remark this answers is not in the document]
vojtechsimetka: Yes please. `fetchBalances` sounds better.

src/lib/objects/adapter.ts
```diff
@@ -14,12 +24,73 @@ export function makeWakuObjectAdapter(adapter: Adapter, wallet: BaseWallet): Wak
 		return adapter.estimateTransaction(wallet, to, token)
 	}
 
-	async function checkBalance(token: Token): Promise<void> {
-		await adapter.checkBalance(address, token)
+	async function _checkBalance(token: Token): Promise<void> {
+		await checkBalance(address, token)
+	}
+
```
vojtechsimetka, line 30: Why does the object check balances? Shouldn't the app be just checking the balances onload and maybe periodically?

vojtechsimetka (changes requested):
Please rename the `initializeBalances` to `fetchBalances`.

I would also love to keep the `firebase` adapter do everything just in the `firebase` db (including all transactions etc). I am fine to merge this PR but then I would love to restore the way it originally worked just as we discussed.

src/lib/objects/adapter.ts
agazso: It's because the balance is get through polling and with polling there is always a tradeoff that you either make too many polls and waste resources or too few and then it may be a poor user experience.

I think we should check the balance periodically (let's say once in every minute or so) but if there is a transaction done and it can be expected that changes the balance then it would be good to keep to possibility to force a `checkBalance`.

vojtechsimetka (approved):

Issue #183, opened 2023-07-12
Chat view is 100% tall, scrolling looks wrong on mobile


agazso:
Because the chat view is 100% of the viewport height, when scrolling the scrollbar is also visible on the full height of the app, going over elements which are not vertically scrollable themselves, like the header and the input fields on the bottom.

It would be better if the scrollbar only appeared over the chat view itself and not over other non-scrollable elements.

![Screenshot from 2023-07-12 17-26-47](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/3eca74a6-25e6-4526-bddb-321ba29d74db)

PR #184, opened 2023-07-13
style: fix chat screen scrollbar

sugarpusher:
Solves #183

### DIFF
--- a/src/routes/chat/[id]/+page.svelte
+++ b/src/routes/chat/[id]/+page.svelte
@@ -28,26 +28,19 @@
 	let autoscroll = true
 
 	beforeUpdate(() => {
-		if (browser && window && div) {
-			const height = div.scrollHeight
-			autoscroll = height <= window.scrollY + window.innerHeight
-		}
+		autoscroll = div && div.offsetHeight + div.scrollTop > div.scrollHeight - 74
 	})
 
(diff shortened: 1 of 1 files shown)


agazso (commented):
This fix solves the scrollbar issue, so that's good news. However it causes a regression in that it no longer scrolls to the last message in case when there is more text in the chat than one screen.

sugarpusher:
Thanks for catching @agazso 
Please re-review, it seems to work well for me now.
I used this as reference: https://svelte.dev/tutorial/update
Thanks!

agazso (changes requested):
For some reason I don't see the [2nd commit that fixes the scroll](https://github.com/logos-innovation-lab/waku-objects-playground/commit/cad5a56913835a186168492fd92533b1ef766d8a) on github in the PR view, so I cannot comment on that directly.

So the functionality works now and it scrolls to the last message. The only difference is that previously the app scrolled smoothly. In order to fix that change the code in `afterUpdate` from
`if (autoscroll) div.scrollTo(0, div.scrollHeight)`
to
`if (autoscroll) div.scrollTo({ top: div.scrollHeight, behavior: 'smooth' })`

Then it will work exactly like before.

agazso (approved):

PR #202, opened 2023-07-26
feat: dark mode color switching

vojtechsimetka:
IMHO we can merge this and then tackle the actual dark version of colors in individual components as a separate task PR


sugarpusher (approved):

sugarpusher:
I agree.

PR #205, opened 2023-07-26
style: dark mode selected from app (not browser)

sugarpusher:


src/lib/components/select.svelte
```diff
@@ -37,12 +37,12 @@
 		font-weight: 600;
 		font-size: var(--font-size-normal);
 		white-space: nowrap;
-		color: var(--color-body-text);
+		// color: var(--color-body-text);
```
vojtechsimetka, line 40: Is this commented out intentional? Can we remove?

```suggestion
```

vojtechsimetka (approved):
LGTM thanks!

PR #212, opened 2023-07-27
chore: remove double border on Identity screen

sugarpusher:


agazso (commented):
Nice!

Maybe that's a different issue but it seems there is also double border on the assets on the account page.

vojtechsimetka:
@agazso I don't actually see it. Can you create issue with a screnshot please?

vojtechsimetka (approved):

agazso:
> @agazso I don't actually see it. Can you create issue with a screnshot please?

I opened an issue: #219

PR #215, opened 2023-07-27
fix: transaction status update

### LINKED ISSUE #187: Payggy transaction status update
Currently there are two notifications for a transaction: once when the transaction is initiated, once when the payment is succesful. The reason for this is that there are two waku data messages (the first for the pending tx, the second for the successful one). 

It would be better to just show one instead, by sending the initial tx message and update the state locally.

Screenshot from 2023-07-19 16-43-43

agazso:
This is a fix for avoiding two widgets in the chat when sending a transaction. It works by waiting for the transaction state change when the message arrives and updating the object's store when that happens.

An edge case where this would not work is when the app is reloaded when the transaction is in pending state. If that happens the transaction will forever stay in pending state. A solution to fix this would be some kind of internal queue for pending transactions.

Fixes #187

### DIFF
--- a/src/lib/adapters/waku/index.ts
+++ b/src/lib/adapters/waku/index.ts
@@ -87,8 +87,8 @@ async function executeOnMessage(address: string, adapter: WakuObjectAdapter, cha
 
 		if (descriptor && descriptor.onMessage && wakuObjectStore.lastUpdated < chatMessage.timestamp) {
 			const objects = wakuObjectStore.objects
-			const store = objects.get(key)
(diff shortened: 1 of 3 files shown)


vojtechsimetka (approved):
LGTM. I don't understand why this would not work after refresh of the page. Is the `onMessage` not called after refresh?

agazso:
> LGTM. I don't understand why this would not work after refresh of the page. Is the `onMessage` not called after refresh?

You are right, the `onMessage` is called after refresh and this will work correctly.

PR #218, opened 2023-07-27
fix: number conversion exception

### LINKED ISSUE #182: Number conversion with `toBigInt` causes unhandled exception
There are several bigint conversions on the Payggy standalone page, for example when entering the amount. If an invalid number is entered it causes an unhandled exception and that breaks the whole app, for example the navigation stops working after.

This can be traced back to the `toBigInt` calls and the `validateDecimal` calls. Even though the return type of `toBigInt` correctly indicates that it can throw an exception with the `never` type, this is not enforced by the compiler and there are no warnings if the caller ignores it.

One solution could be to change the return type to `bigint | Error`, therefore enforce the caller to properly handle if the amount is invalid. If we do this then we can also change `Input` type to accept the error type and display it correctly to the user, therefore simplifying the error handling case.

agazso:
Fixes #182

### DIFF
--- a/src/lib/objects/payggy/standalone.svelte
+++ b/src/lib/objects/payggy/standalone.svelte
@@ -54,8 +54,12 @@
 	let toUser =
 		args.users.find((user) => user.address !== args.profile.address) || throwError('invalid user')
 	$: if (amount && token) {
-		const tokenToTransfer = { ...token, amount: toBigInt(amount, token.decimals) }
-		args.estimateTransaction(toUser.address, tokenToTransfer).then((f) => (fee = f))
+		try {
+			const tokenToTransfer = { ...token, amount: toBigInt(amount, token.decimals) }
+			args.estimateTransaction(toUser.address, tokenToTransfer).then((f) => (fee = f))
+		} catch (e) {
+			console.log({ e })
+		}
 	}
 	// $: if (!amount && store && store.view === 'overview') history.back()
 


src/lib/objects/payggy/standalone.svelte
```diff
@@ -54,8 +54,12 @@
...
 		args.users.find((user) => user.address !== args.profile.address) || throwError('invalid user')
 	$: if (amount && token) {
-		const tokenToTransfer = { ...token, amount: toBigInt(amount, token.decimals) }
-		args.estimateTransaction(toUser.address, tokenToTransfer).then((f) => (fee = f))
+		try {
+			const tokenToTransfer = { ...token, amount: toBigInt(amount, token.decimals) }
+			args.estimateTransaction(toUser.address, tokenToTransfer).then((f) => (fee = f))
+		} catch (e) {
+			console.log({ e })
+		}
```
vojtechsimetka, line 62: I agree we can do this for now, but the solution is to have `amountString` and `amountBigInt` which gets parsed here and when there is error, we need to set the `fee` to `0` and the `amountBigInt` to `0` as well.

vojtechsimetka (approved):

src/lib/objects/payggy/standalone.svelte
agazso: I can also imagine that the input does not allow entering an invalid value. I tried to make the input accept the `type` passed in as prop and use a numeric input but unfortunately svelte does not allow this when the value is bound.

Issue #219, opened 2023-07-28
Separators are not consistent between screens


agazso:
The separators on the Identity screen has been changed recently (#212) so that it is thinner. It looks nice now, but there are similar separators on the account page or the contacts and they look thicker. See the attached screenshots.

![Screenshot from 2023-07-28 10-40-12](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/8261172d-1250-4896-a825-32bf54a42449)

![Screenshot from 2023-07-28 10-42-01](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/4349f153-64c0-47bc-96eb-51cb893010c5)

sugarpusher:
@agazso I think they're darker, not thicker - but I'll take a look. Thanks!

PR #220, opened 2023-07-28
style: fix border color on main css file variable

sugarpusher:
Solves #219


agazso (approved):
Thank you, looks much better!

PR #223, opened 2023-07-28
feat: remove firebase adapter

### LINKED ISSUE #221: Remove firebase adapter

agazso:
Fixes #221

This may break an existing instance if it used firebase previously. In that case the `adapter` field should be deleted (or the value should be changed to `waku`) in the local storage in the dev tools.


vojtechsimetka:
> This may break an existing instance if it used firebase previously. In that case the `adapter` field should be deleted (or the value should be changed to `waku`) in the local storage in the dev tools.


Since we don't have any other adapters, I suggest to just export the waku adapter as defaults with `export default new WakuAdapter()`.
Screenshot 2023-07-28 at 16 13 51

vojtechsimetka (approved):
lGTM thanks for doing this!

Issue #214, opened 2023-07-27
The Payggy widget width in the chat depends on the amount


agazso:
The width of the widget depends on the amount provided and that does not conform to the grid. Also there is space next to the amount and the text is broken into two lines so it looks a bit off.

![Screenshot from 2023-07-27 16-51-57](https://github.com/logos-innovation-lab/waku-objects-playground/assets/230163/909c8c84-532b-4054-83b3-e0e8c3399602)

sugarpusher:
@agazso On 7/19 I mentioned on the dev channel that I was getting a `Failed to parse store or message data. Check console for details.` error on the Payggy bubbles and a `ZodError` in the console. 
@vojtechsimetka mentioned that the old Payggy objects didn't work with the new changes and also that I needed xDAi to make transactions. I'm still getting those errors, is it because I don't have xDai? How do I get some?
Thanks!

agazso:
> @agazso On 7/19 I mentioned on the dev channel that I was getting a `Failed to parse store or message data. Check console for details.` error on the Payggy bubbles and a `ZodError` in the console. 

Opened an issue about it:
https://github.com/logos-innovation-lab/waku-objects-playground/issues/216

> @vojtechsimetka mentioned that the old Payggy objects didn't work with the new changes and also that I needed xDAi to make transactions. I'm still getting those errors, is it because I don't have xDai? How do I get some? Thanks!

This worked for me https://www.gnosisfaucet.com/

PR #224, opened 2023-07-28
style: make Payggy bubble fullwidth

sugarpusher:
Payggy bubble is now always fullwidth.
I added a `fullwidth` prop to the ChatMessage component so we can specify per waku object.

Solves #214

### DIFF
--- a/src/lib/components/chat-message.svelte
+++ b/src/lib/components/chat-message.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	export let bubble = false
 	export let myMessage = false
+	export let fullwidth = false
 
 	const isFF = () => {
 		let browserInfo = navigator.userAgent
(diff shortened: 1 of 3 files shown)


agazso (approved):
I tried it and it looks good.

I don't think we need the `fullwidth` property and the ability to set it per object. I think by design the waku objects are always full width (that is 3/4 of the width of the chat). Maybe ask @david-gauquelin about it?

vojtechsimetka (approved):

vojtechsimetka:
I also think everything is fullwidth always.

vojtechsimetka:
So if I understand it correctly, @sugarpusher we should remove the `fullwidth` param and just make it fullwidth by default

vojtechsimetka (approved):

agazso (commented):
Looks good!

PR #225, opened 2023-07-30
fix: description and viewport in app.html

michelleplur:
Added viewport / disable zooming /doubletap zoom


agazso:
There is a problem with linting. Can you please run `pnpm format` and commit and push to this branch to fix it?

PR #228, opened 2023-07-31
feat: group chat

agazso:
This is still WIP.

Open questions and issues:
- [ ] shall we use the same `Chat` type and the same `chats` map in the chat store (just differentiate with a field or something) or shall we use different types/objects? Both has advantages and disadvantages.
- [ ] currently the contacts are not maintained directly, however it is possible to obtain a list of contacts from the users of the private chats. 
- [ ] the current dataflow model with the `WakuAdapter` is that the app invokes an adapter function, that does something and as a side effect it updates the state store, which in turn updates the UI. This makes the state store effectively read-only from the UI perspective (e.g. no updates are permitted) because otherwise there is no guarantee that the state store is persisted.
- [ ] the actual group chat implementation can either use 1-to-N messaging or just a common, shared topic of the group. The 1-to-N messaging is used in real-life cryptographic implementations, however with a naive implementation it is more likely that the group chat ends up in an inconsistent state. Because of this I tend to think of using the shared topic approach.


vojtechsimetka (approved):
Few bugs I encountered:
- [ ] if I do not have someone from the group in my contacts, I don't become part of the group
- [ ] if I am offline and someone invites me to a group I don't become part of the group
- [ ] I can not change the name of the group
- [ ] payggy should be disabled in group if there is more than 2 users in total (the recipient is kinda selected at random in this case). Or we should allow the sender to choose from the group members who should be the recipient (I quite like this).

May not be a bug:
- [ ] I can create group with just two members (basically this becomes direct chat)

agazso:
> Few bugs I encountered:
> 
>     * [ ]  if I do not have someone from the group in my contacts, I don't become part of the group

I could not reproduce this, maybe one of the users were offline?

>     * [ ]  if I am offline and someone invites me to a group I don't become part of the group

Yeah, this is definitely an issue.

>     * [ ]  I can not change the name of the group

Yeah, this too.

>     * [ ]  payggy should be disabled in group if there is more than 2 users in total (the recipient is kinda selected at random in this case). Or we should allow the sender to choose from the group members who should be the recipient (I quite like this).

Yeah, this was also my last point in the description.

> May not be a bug:
> 
>     * [ ]  I can create group with just two members (basically this becomes direct chat)

This is not a bug, because you can add members later to this chat. You can not add members to the direct chat.

I will open issues about these.

PR #244, opened 2023-08-03
fix: stuck on loading when getting an invite without existing identity

### LINKED ISSUE #243: Invite screen shows the loading screen indefinitely
> There's an error when the app tries read from the browser's localstorage.
Not sure if related. 

https://discord.com/channels/973324189794697286/1133963048806268951/1136559778752102400

Screenshot from 2023-08-03 10-29-43

vojtechsimetka:
Resolves #243

### DIFF
--- a/src/routes/invite/[address]/+page.svelte
+++ b/src/routes/invite/[address]/+page.svelte
@@ -105,7 +105,7 @@
 	</Button>
 </Header>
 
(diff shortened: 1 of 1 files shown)


agazso (approved):
This fixes the issue, but I wonder is there a way to handle these problems better so that it is less fragile in the future.

vojtechsimetka:
Yeah that is exactly my thought as well.

PR #254, opened 2023-08-03
feat: introduce unread counter

vojtechsimetka:


sugarpusher (approved):

agazso:
It would be better to add an optional field to the `UserMessage` type:
`read?: true` 

Then it could be counted with `const numUnread = messages.filter(message => !message.read).length`

vojtechsimetka:
We can implement it this way in the adapter. Still, in the store I would love to see just one number that gets displayed. I want the stores to be as simple as possible.

PR #256, opened 2023-08-03
fix: remove unused all users contacts

### LINKED ISSUE #246: After accepting the invite the peer is always undefined

agazso:
Deleted the `all-users` store from waku and also the contacts state store. We no longer need any of these, but they caused issues because of duplication. We can simply use the shared `profile` on waku and the `contacts` can be calculated from the `chats`.

Fixes #246


vojtechsimetka (commented):
I very much like the change, but it does not work for me. I still don't see these users

`0xdF4b5ed27Bd2B4774288f09A4A5Ab0033491d9A8`
`0xb4bee9071Dd82aEdF3AA7d18c0d2a3eD3Bd22EDf`

I am pretty sure that at least the first on has profile setup
Screenshot 2023-08-03 at 17 16 58

agazso:
> I very much like the change, but it does not work for me. I still don't see these users

I pushed a change to check if the profiles has changed on startup. Can you please test it again?

vojtechsimetka (approved):

PR #276, opened 2023-08-07
fix: 500 error when reloading waku object

### LINKED ISSUE #178: Reloading standalone page fails with 500 error
When reloading the standalone page (e.g. `/chat/{address}/object/{objectId}/new`) a 500 Internal Error is reported. On the console log there is a message 'no wallet`.

### LINKED ISSUE #274: Reloading the /chat../object/... pages results in 500 Internal error

vojtechsimetka:
Fixes #274
Fixes #178

TODO:
- [x] extract create identity into a component
- [x] create a wrapper or a mechanism that checks if user has wallet on every page

Would love to tackle separately:
- [ ] handle the case where the chat does not exist or user can not view it


sugarpusher (approved):
LGTM

PR #284, opened 2023-08-08
fix: ReadonlyText component spacings

sugarpusher:
Updates ReadonlyText component (and pages that use it) and fixes jaggedness of Payggy detail open/close animation


agazso (approved):
It looks good. The only thing I found is that when the details are not open the `Details` label should be not bold. I found the change of the size of the text strange, but that seems to be by design.

sugarpusher:
> 

Cool, thanks for catching the bold. I'll fix it. Thanks!

Issue #241, opened 2023-08-02
Scrollable views are often 100% tall, scrolling looks weird


agazso:
The easiest to find out is to use the mobile view and choose a phone which has smaller screen (e.g. iPhone SE).

The identity screen has definitely this issue and I suspect others as well (e.g. I think the account screen if there are more currencies). I think it would be good to fix this problem in a systematic way, rather than fixing individual screen.

sugarpusher:
@agazso Yup you're correct. I'll go over it globally. Thanks!

PR #288, opened 2023-08-09
fix: global scrolling issue

sugarpusher:
Note: This PR doesn't include any changes to the group chat screens - they will be addressed separately, as the styling hasn't been finished for them

Solves #241

### DIFF
--- a/src/lib/components/authenticated-only.svelte
+++ b/src/lib/components/authenticated-only.svelte
@@ -26,17 +26,17 @@
 </script>
 
 {#if $profile.loading || $walletStore.loading}
-	<Container align="center" grow gap={6} justify="center">
+	<Container align="center" gap={6} justify="center">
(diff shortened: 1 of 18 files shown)


agazso (approved):
It works well on the screens that I tested.

I found one small issue, that can be addressed in a separate PR though:
- [ ] Identity page missing padding between bottom and `Disconnect...` button

sugarpusher:
> It works well on the screens that I tested.
> 
> I found one small issue, that can be addressed in a separate PR though:
> 
> * [x]  Identity page missing padding between bottom and `Disconnect...` button

This was easy enough so I went ahead and did it. There was already a Spacer component there but for some reason the height wasn't working. I switched to use padding instead.

vojtechsimetka (approved):

PR #289, opened 2023-08-09
fix: payggy default token balance

### LINKED ISSUE #286: When sending tokens (Payggy), the xDai balance shows as 0 until after re-selecting the token from the dropdown

agazso:
Fixes #286

### DIFF
--- a/src/lib/objects/payggy/standalone.svelte
+++ b/src/lib/objects/payggy/standalone.svelte
@@ -32,7 +32,6 @@
 		type SendTransactionDataMessage,
 	} from './schemas'
 	import type { Token } from '../schemas'
-	import { defaultBlockchainNetwork } from '$lib/adapters/transaction'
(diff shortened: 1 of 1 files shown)


src/lib/objects/payggy/standalone.svelte
```diff
@@ -54,9 +53,8 @@
 	let token: Token
 	let amount = ''
 	let fee: Token | undefined = undefined
-	const nativeToken = defaultBlockchainNetwork.nativeToken
 
-	$: if (!token) token = nativeToken
+	$: if (!token) token = args.tokens[0]
```
vojtechsimetka, line 57:
```suggestion
	$: if (!token) token = args.tokens.find(t => t.address === undefined) ?? tokens[0]
```

vojtechsimetka (approved):
I'd prefer to always have native token to be selected. Otherwise LGTM

PR #291, opened 2023-08-09
feat: link support in chat

agazso:


sugarpusher:
I noticed that unless I start with `http://`, the link doesn't work - it thinks I'm looking for a relative path and I get a "Chat not found" error. Maybe we can make it work when starting the link with `www.`?

agazso:
> I noticed that unless I start with `http://`, the link doesn't work - it thinks I'm looking for a relative path and I get a "Chat not found" error. Maybe we can make it work when starting the link with `www.`?

Thanks for catching this. I added a library that should handle all cases and also some tests.

vojtechsimetka (approved):
nice, works great! :)

agazso:
Because of the errors in the preview closing this and using #293 instead

PR #296, opened 2023-08-14
style: group chat

sugarpusher:


agazso (approved):
In general it looks good. I have two comments:
- Please change the title of the PR to `style: ...` instead of `wip: ...`
- The default avatar icon in the group chat makes the text bubble look a bit weird, maybe it would be good if @david-gauquelin could take a look at it

Screenshot from 2023-08-14 16-37-18

david-gauquelin:
Three steps to fix this (from minimal to best):
1. Let's change the background color of the default avatar image to #E7E7E7
2. Let's use dynamic colors for the avatar (as set in the appearance), then the avatar icon should be set to "ultraHigh" (5% luminance and the avatar background to "low" (80% luminance)
3. Let's use https://www.dicebear.com/ to generate random avatar images (use the style "Bottts neutral) as shown in the design mockups on Sketch

PR #306, opened 2023-08-15
refactor: chat store helpers for simplifying common functionality

agazso:


sugarpusher (approved):

src/lib/stores/chat.ts
```diff
@@ -60,7 +64,57 @@ function createChatStore(): ChatStore {
...
+				if (!state.chats.has(chatId)) {
+					return state
+				}
+				const newChats = new Map<string, Chat>(state.chats)
+				newChats.delete(chatId)
+
+				return {
+					...state,
+					chats: newChats,
+				}
```
vojtechsimetka, line 114: There is no need to go through the `newChats` ordeal. You can modify the original `state.chats` object and just return it. It is more efficient as it does not create new objects.

```suggestion
				state.chats.set(chat.chatId, chat)

				return {
					...state,
					chats: state.chats,
					loading: false,
				}
			})
		},
		updateChat: (chatId: string, update: (chat: Chat) => Chat) => {
			store.update((state) => {
				if (!state.chats.has(chatId)) {
					return state
				}
				const oldChat = state.chats.get(chatId)
				if (!oldChat) {
					return state
				}
				state.chats.set(chatId, update(oldChat))

				return {
					...state,
					chats: state.chats,
				}
			})
		},
		removeChat: (chatId: string) => {
			store.update((state) => {
				if (!state.chats.has(chatId)) {
					return state
				}
				state.chats.delete(chatId)

				return {
					...state,
					chats: state.chats,
				}
```

vojtechsimetka (approved):
Thanks for making this bit simpler!

Issue #128, opened 2023-06-28
Payggy: implement pricing in fiat


vojtechsimetka:
Provide a mechanism to get the price of token asset. We need both current and at the time of transaction (which can be stored somewhere).

PR #310, opened 2023-08-17
feat: split object

### LINKED ISSUE #350: payggy object keeps in loading state on receiver side
Screenshot 2023-09-06 at 16 31 12

vojtechsimetka:
Resolves #350

TODO:
- [x] find out why the object is not processing onMessage calls
- missing ui screens
    - [x] activity screen
    - [x] collection screen
    - [x] expenses screen
    - [x] settle up screen
- integration with the contracts
    - [x] create new splitter contract
    - [x] add expense to existing splitter
    - [x] fetch balances from splitter
    - [x] settle
- [x] user input validation

Tackle separately:
- use native token symbol everywhere rather than hardcoded DAI
- use `toSignificant` everywhere to show precision numbers consistently
- on startup fetch latest data from contract
- estimate transaction cost #314
- show correctly DAI ("fiat") exchange amount #128


src/lib/adapters/waku/index.ts
```diff
@@ -140,6 +141,7 @@ async function executeOnDataMessage(
 		const users = chat ? chat.users : []
 		const args: WakuObjectArgs = {
 			...context,
+			chatName: chat?.name ?? 'Unknown', // FIXME: figure out how to have a proper chatname
```
agazso, line 144: `chatName` is either `chat.name` (for group chats) or it is the name of person you talk to in a private chat (that is `users[0].name` by convention)

src/lib/objects/index.d.ts
```diff
@@ -31,22 +31,30 @@ export interface WakuObjectState {
...
 }
 
-type StoreType = JSONSerializable
-type DataMessageType = JSONSerializable
-
-export interface WakuObjectContext extends WakuObjectAdapter {
+export interface WakuObjectContext<
+	StoreType extends JSONSerializable = JSONSerializable,
+	DataMessageType extends JSONSerializable = JSONSerializable,
+	ViewType extends string = string,
```
agazso, line 40: What is the reason for making this generic?

vojtechsimetka: For a type safety. For example the `onViewChange` function now only allows specific string literals or the store is properly typed. Note that providing types is optional.

src/lib/adapters/waku/index.ts
vojtechsimetka: Oh nice, can I rely on the `[0]` index? Or should I use `users.find(u => u.address !== profile.address)?.name`

src/lib/objects/index.d.ts
```diff
@@ -31,22 +31,30 @@ export interface WakuObjectState {
...
+> extends WakuObjectAdapter {
 	readonly store?: StoreType
 	updateStore: (updater: (state?: StoreType) => StoreType) => void
 
 	send: (data: DataMessageType) => Promise<void>
 
-	readonly view?: string
-	onViewChange: (view: string) => void
+	readonly view?: ViewType // Screen to show on Waku object
+	readonly params: string[] // Other path params after the screen
```
agazso, line 48: If I understand correctly this is used so that the individual screens can be addressed directly with URLs.

If that's the case I would rather call this `viewParams` because otherwise the naming is not very clear, or at least for me it is puzzling when I see `args.params` in the code.

src/lib/objects/split/blockchain.ts
```diff
@@ -0,0 +1,106 @@
...
+	// TODO: Maybe we don't need metadata
+	const tx = await splitterFactory.create(metadata, token, members)
+	const receipt = await tx.wait()
+
+	if (!receipt || receipt.status !== 1) {
+		throw new Error('Failed to create splitter contract')
+	}
+
+	let splitterAddress: string | undefined = undefined
+	for (let sleepTime = 100; sleepTime < 2000 && splitterAddress === undefined; sleepTime += 100) {
```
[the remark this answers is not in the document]
vojtechsimetka: That was my first approach but it did not really work. I'll give it another shot

src/lib/objects/split/index.ts
```diff
@@ -0,0 +1,98 @@
...
+			console.error(res.error)
+			return
+		}
+
+		if (res.data.type === 'expense') {
+			const { expense, splitterAddress } = res.data
+
+			let balances: Balance[] = []
+			try {
+				balances = await getBalances(args.getContract, splitterAddress)
```
agazso, line 34: Under what circumstances this may fail? This looks like the error is used as a regular conditional, but I suspect that it may have multiple reasons and maybe it would make sense to more clearly differentiate. For example if the blockchain connection fails maybe we should not calculate things and update the store.

vojtechsimetka: You are right and I actually wanted to discuss this with you. It may fail because the contract address is wrong or because the blockchain RPC connection fails. Normally I would just leave the `balances = []` but I already had the offchain function to calculate ready.

I guess we should have a broader discussion about error handling. Not only blockchain error but also waku errors or maybe even attacks. For example sometimes even object messages arrives twice e.g. new expense which happened few times during development but no idea how to reproduce. I'll create an issue about it.

src/lib/adapters/waku/index.ts
agazso: Currently we rely on the `[0]` index elsewhere in the code, but that is not really documented anywhere and it's not enforced either, so it may be better to use the `find` version.

It would be nice to add this to a utility function in the chat store, so if we change it later then we only need to change it at one place. This can be done in a separate PR though.

src/lib/objects/split/blockchain.ts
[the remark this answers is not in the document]
agazso: If it results in a more convoluted solution an alternative could be to use constants to clearly define what is happening. 
```suggestion
        const sleepTimeout = 2000
        const sleepTime = 100
	for (let elapsedTime = 0; elapsedTime < sleepTimeout && splitterAddress === undefined; elapsedTime += sleepTime) {
```

agazso (approved):
Nice, tested the flows, worked well. After testing it I would iterate on some of the design decisions too, but that does not affect this PR.

Some additional comments, some of it may be styling related:
- The widget says `unknown shared expenses` I am not sure where this can be changed, if it cannot be changed then it's better to not display it at all.
- In the Send transaction list screen: `You owes XXX` (should be `You owe XXX`)
- If you enter invalid amount the on the initial screen of the Split the form moves up to print an error. It would be better if the form did not move and the error would be displayed under the form.

PR #311, opened 2023-08-17
feat: group bubbles from same sender

sugarpusher:
And only show name and avatar on last bubble

### DIFF
--- a/src/routes/group/chat/[id]/+page.svelte
+++ b/src/routes/group/chat/[id]/+page.svelte
@@ -167,7 +167,7 @@
 										{@const lastMessage =
 											i + 1 === messages.length ||
 											messages[i].fromAddress !== messages[i + 1]?.fromAddress ||
-											messages[(i + 1) % len].type !== 'user'}
+											messages[i + 1]?.type !== 'user'}
 										{@const sender = chat.users.find((u) => message.fromAddress === u.address)}
 										<ChatMessage
 											myMessage={message.fromAddress === wallet.address ? true : false}


src/lib/components/chat-message.svelte
```diff
@@ -3,6 +3,7 @@
 	export let myMessage = false
 	export let object = false
 	export let group = false
+	export let sameSender = false
 	export let sender: undefined | string = undefined
```
vojtechsimetka, line 7: I am starting to feel like there is a lot of flags. Can you add comment to them what they indicate? I am starting to get lost.

src/routes/group/chat/[id]/+page.svelte
```diff
@@ -160,18 +160,29 @@
...
 								<!-- Chat bubbles -->
-								{#each messages as message}
+								{#each messages as message, i}
 									{#if message.type === 'user' && message.text.length > 0}
+										{@const len = messages.length}
+										{@const sameSender =
+											messages[i].fromAddress === messages[(i + len - 1) % len].fromAddress}
+										{@const lastMessage =
+											messages[i].fromAddress !== messages[(i + 1) % len].fromAddress ||
+											messages[(i + 1) % len].type !== 'user'}
```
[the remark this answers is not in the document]
Comment (sugarpusher) [reply src/routes/group/chat/[id]/+page.svelte:170]: @vojtechsimetka 
I am using these to group messages from the same person in a group chat. The first and last messages have the curvy top, the middle ones don't. Also, the name and avatar should only show on the last message of the group -- except for my own messages. Those never show the name or avatar.

I see the issue you're talking about, thank you for flagging it. 
If you have any suggestions, I'm open 😊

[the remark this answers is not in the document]
Comment (agazso) [reply src/routes/group/chat/[id]/+page.svelte]: The modulo is indeed not needed and can cause logical errors. You can simply use `messages[i].fromAddress === messages[i + 1]?.fromAddress` because if you go over the last index then the indexing operator will return `undefined`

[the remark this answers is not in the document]
Comment (sugarpusher) [reply src/routes/group/chat/[id]/+page.svelte]: @agazso 
man that was my first attempt but I forgot the `?` and kept getting undefined errors so I gave up and was trying something else 🤦‍♀️
Thank you!

src/routes/group/chat/[id]/+page.svelte
```diff
@@ -160,18 +160,27 @@
...
 								<!-- Chat bubbles -->
-								{#each messages as message}
+								{#each messages as message, i}
 									{#if message.type === 'user' && message.text.length > 0}
+										{@const len = messages.length}
+										{@const sameSender = messages[i].fromAddress === messages[i - 1]?.fromAddress}
+										{@const lastMessage =
+											i + 1 === messages.length ||
+											messages[i].fromAddress !== messages[i + 1]?.fromAddress ||
+											messages[(i + 1) % len].type !== 'user'}
```
Comment (agazso) [inline src/routes/group/chat/[id]/+page.svelte:170]: There is still modulo here which is incorrect. Try this instead:
```suggestion
											messages[i + 1]?.type !== 'user'}
```

Comment (sugarpusher) [reply src/routes/group/chat/[id]/+page.svelte:170]: Sorry and thank you. I have a pretty bad headache today so I'm not really thinking straight. 😞

Comment (agazso) [reply src/routes/group/chat/[id]/+page.svelte]: Get better soon!

agazso (approved):
On Firefox the bubbles are not grouped together, but the avatar is only shown once per block, so it's still okay. I will open a low-priority issue about it.

Otherwise looks good!

PR #313, opened 2023-08-18
refactor: waku storage

### LINKED ISSUE #255: Implement unread messages counter
More info here https://github.com/logos-innovation-lab/waku-objects-playground/pull/254

### LINKED ISSUE #268: Can not change the name of the group
_Originally posted by @vojtechsimetka in https://github.com/logos-innovation-lab/waku-objects-playground/pull/228#pullrequestreview-1563048288_

### LINKED ISSUE #294: Fetch messages I received when I was offline

### LINKED ISSUE #295: Member count is inconsistent between group chat members
Screenshot 2023-08-11 at 11 33 04 AM
Screenshot 2023-08-11 at 11 32 36 AM

### LINKED ISSUE #309: Solve invite workaround so that the Decline invite functionality works

agazso:
This is quite a major change to address many previous issues regarding reliability and consistency. Introduces a `Wakustore` type that is loosely modeled after the `Firestore` API from Firebase. There are also new types introduced to differentiate between what is stored on the waku network and what is the local svelte store.

- There is proper history management for the private and group messages and it is based on the implementation of status-web that uses the `timestamp` of the waku messages and time based queries. (Fixes #294 and #309) Previously the `Message` type also defined a timestamp, but that is set by the sender and not the node, so it is not suitable for querying. Because of this there is a hack to overwrite the `Message` timestamp with the waku message's timestamp.
- Added calculation of the unread counter and set it to zero when the user opens the chat. (Fixes #255)
- There are proper subscriptions now for private and group messages but also for the group chat itself to automatically update the group chat state when a new member is added to the group (Fixes #295 and #268)

The caveat is that the data flow is still not strictly unidirectional:
- The `chats`, `profile` and `objects` local stores are written locally and then saved to the waku network
- The `group-chats` are written to the network first and then it updates the local store in the listener callback
- The `private-message` is a mix that it's written to the network and listened to and that updates the `chats` local store

Fixes #255 
Fixes #268
Fixes #294
Fixes #295
Fixes #309

### DIFF
--- a/src/lib/adapters/index.ts
+++ b/src/lib/adapters/index.ts
@@ -14,6 +14,7 @@ export interface Adapter {
 	startGroupChat(wallet: BaseWallet, chat: DraftChat): Promise<string>
 
 	addMemberToGroupChat(chatId: string, users: string[]): Promise<void>
+	removeFromGroupChat(chatId: string, address: string): Promise<void>
 	saveGroupChatProfile(chatId: string): Promise<void>
 
 	sendChatMessage(wallet: BaseWallet, chatId: string, text: string): Promise<void>
--- a/src/lib/adapters/index.ts
+++ b/src/lib/adapters/index.ts
@@ -10,7 +10,7 @@ export interface Adapter {
 	saveUserProfile(address: string, name?: string, avatar?: string): Promise<void>
 	getUserProfile(address: string): Promise<User | undefined>
 
-	startChat(address: string, chat: DraftChat): Promise<string>
+	startChat(address: string, peerAddress: string): Promise<string>
 	startGroupChat(wallet: BaseWallet, chat: DraftChat): Promise<string>
 
 	addMemberToGroupChat(chatId: string, users: string[]): Promise<void>
--- a/src/lib/adapters/index.ts
+++ b/src/lib/adapters/index.ts
@@ -14,6 +14,7 @@ export interface Adapter {
 	startGroupChat(wallet: BaseWallet, chat: DraftChat): Promise<string>
 
 	addMemberToGroupChat(chatId: string, users: string[]): Promise<void>
+	saveGroupChatProfile(chatId: string): Promise<void>
 
 	sendChatMessage(wallet: BaseWallet, chatId: string, text: string): Promise<void>
 	sendData(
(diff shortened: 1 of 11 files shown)


vojtechsimetka (changes requested):
Wow great job, it is much cleaner and makes much more sense!

Critical issues
- I can not create new identity, when I try to save my username it errors with `invalid user to save: 0xd8f57337dCd1D5B7a6b86b297db9d83041DB55f9` . So I remain as undefined and noone can start a chat with me. I can seemingly start chat with others, but they don't see the chat.

Nice to fix, can be done in separate PRs
- `src/lib/routes/group/chat/[id]/+page.svelte` The `decline` function is not hooked up so not sure whether #309 is fixed.
- the unread counter does not work if user is online and has the chat open. So `A` chats with `B`. Both have the chat open, `A` sends message to `B`. `B` sees the message in the chat (it scrolls to the message). `B` goes to all chats where the chat with `A` says there are unread messages.

agazso:
>     * I can not create new identity, when I try to save my username it errors with `invalid user to save: 0xd8f57337dCd1D5B7a6b86b297db9d83041DB55f9` . So I remain as undefined and noone can start a chat with me. I can seemingly start chat with others, but they don't see the chat.

This is fixed now.

> Nice to fix, can be done in separate PRs
> 
>     * `src/lib/routes/group/chat/[id]/+page.svelte` The `decline` function is not hooked up so not sure whether [Solve invite workaround so that the Decline invite functionality works #309](https://github.com/logos-innovation-lab/waku-objects-playground/issues/309) is fixed.
> 
>     * the unread counter does not work if user is online and has the chat open. So `A` chats with `B`. Both have the chat open, `A` sends message to `B`. `B` sees the message in the chat (it scrolls to the message). `B` goes to all chats where the chat with `A` says there are unread messages.

I also fixed these, because during the testing I encountered so many errors with the waku storage. After some investigation I found out that there were way too many requests to the waku store and eventually they started failing.

There are two major fixes that were added:
- Saving the local chats to the network is done with rebounce and a counter so that if there is a saving in progress then wait until it is finished and then save again with the updated state
- The message handling was asynchronous but the callback did not wait for finishing it. I added a queue to the message handling.

With these two changes I was able to track down the decline and the unread issues as well.

vojtechsimetka (approved):
Thanks, all the issues were resolved and it feels like it performs way better than before!

PR #318, opened 2023-08-18
style: link color

### LINKED ISSUE #317: Link style is missing
Since the addition of automatic link detection (#293) there are clickable links in the messages. As far as I know there is no design for the links in the chat, so they are rendered with the default blue color. That is at least visible in light mode, but in dark mode it is very difficult to read.

It would be good to have a chat message link style, or just change the links to the color of the text with underline.

agazso:
Fixes #317

### DIFF
--- a/src/routes/styles.css
+++ b/src/routes/styles.css
@@ -129,6 +129,12 @@ h3 {
 	margin: 0;
 }
 
+a {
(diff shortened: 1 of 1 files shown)


sugarpusher (approved):
Looks good to me

PR #322, opened 2023-08-22
chore: enforce store and message types in object descriptor

vojtechsimetka:


src/lib/objects/hello-world/index.ts
```diff
@@ -17,6 +17,9 @@ export const helloWorldDescriptor: WakuObjectDescriptor = {
 	wakuObject: HelloWorld,
 
 	onMessage: async (address, adapter, store, updateStore, message) => {
-		updateStore(() => message.data)
+		if (message.data) {
```
agazso, line 20: If the definition of the `data` property in the `Message` type is changed from `data?: T` to `data: T` then we don't need this check (the original code will work) and it will work everywhere else too, so I would do that instead.

agazso (approved):

PR #325, opened 2023-08-22
fix: JSONSerializable type

agazso:
There was a bug in the type definition of `JSONSerializable` type that was not caught by the review, neither by the Typescript compiler :grin: . The `JSONValue` was the original name of the `JSONSerializable` but after renaming I forgot to update all occurrences.

After fixing that there was a compiler error of using `{ [key: string]: JSONSerializable }` so I replaced it with a `Record` type instead.


vojtechsimetka (approved):
I actually noticed these but thought it is correct type :D

PR #328, opened 2023-08-23
refactor: extract genRanHex to utils

vojtechsimetka:


src/lib/utils/index.ts
```diff
@@ -0,0 +1,4 @@
+// Note: this is not cryptographically secure
+export function genRanHex(size: number) {
```
agazso, line 2: I would use a more descriptive name if we make it a top level shared function.
```suggestion
export function genRandomHex(size: number) {
```

agazso (approved):

PR #329, opened 2023-08-24
fix: undefined store

agazso:
There is a regression caused by the type-safety refactor and it makes Payggy (and any other WO) using the object store to fail. 

This is just a temporary fix to make it work again, however this is still not typesafe and not completely correct, because `store` can be `undefined`, but it is not a valid `JSONSerializable` value. Still it works for now and in a later PR it can be fixed properly.

### DIFF
--- a/src/lib/objects/ui.svelte
+++ b/src/lib/objects/ui.svelte
@@ -53,8 +53,6 @@
 		chat = $chats.chats.get(chatId)
 	}
 
-	$: console.debug({ store, chat, users, userProfile })
-
 	$: if (chat) {
 		users = chat.users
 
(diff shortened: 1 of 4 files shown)


src/lib/objects/ui.svelte
```diff
@@ -53,7 +53,9 @@
 		chat = $chats.chats.get(chatId)
 	}
 
-	$: if (store && chat) {
+	$: console.debug({ store, chat, users, userProfile })
```
vojtechsimetka, line 56:
```suggestion
```

vojtechsimetka (approved):

PR #358, opened 2023-09-07
fix: entry script CSP for Firefox

filoozom:
We should also remove the `'self'` in `script-src` for the Waku Objects, as it's no longer necessary.

Ideally, the entry script would be hashed and added to `src-script` with `sha256-*`, like we do for the embeds, but that doesn't seem to work for some reason? The current solution might be a bit too open and allow Waku Objects to import Playground scripts, which might be problematic.

### DIFF
--- a/src/lib/objects/external/lib.ts
+++ b/src/lib/objects/external/lib.ts
@@ -53,7 +53,7 @@ export const getNPMObject = async (
 		if (!added['script-src']) {
 			added['script-src'] = ''
 		}
-		added['script-src'] += ` ${window.location.origin + script}`
+		added['script-src'] += ` ${script}`
 
 		return {
 			script,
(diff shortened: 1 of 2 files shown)


filoozom:
> Out of curiosity, what kind of potential problems do you have in mind?

In the current scenario probably nothing concrete, but in theory it would be possible for the Chat App (in this case Playground) to have some JavaScript hosted on its domain that would allow for arbitrary execution. For example, 
https://waku-objects-playground.vercel.app/_app/immutable/chunks/close.1436ed61.js could have an `eval`, and can be imported in the frame by importing `https://waku-objects-playground-qr0x2ro8q-waku-objects.vercel.app/_app/immutable/assets/index.d4b589cf.js/../close.1436ed61.js` even with the `https://waku-objects-playground-qr0x2ro8q-waku-objects.vercel.app/_app/immutable/assets/index.d4b589cf.js` CSP.

agazso (approved):
Tested on Firefox and it works correctly.

PR #359, opened 2023-09-07
feat: external object props

agazso:
This is the host side of this [PR](https://github.com/logos-innovation-lab/waku-objects/pull/21). First that PR needs to be reviewed, merged and published to npmjs so that this one can use it as a dependency. Until that this will be broken.

- Added support for reactive `args`, so that when they change in the host application (e.g. when there is a new user added to a group chat) then a message is sent to the external object to update the arguments and possibly re-render. There is some optimization so that the hash of the serialized `args` is stored on the host side, and an update is only sent if that changes.
- There is also a new message called `init` that the object sends to the host app when the object initialization is complete, so that the host app can send the updated `args` (which is the `context` + `state`). Without this the initialization is full with race-conditions and it is not sure when the object is initialized.
- There is initial support for standalone components, that is when the object is not a chat widget, but it is used when the object needs some initial parameters to be set up or displaying the details of the object. Currently a flag needs to be set in the `getExternalDescriptor`, this could ultimately come from the metadata of the object or alternatively objects could have different targets for chat widget and standalone components. The current implementation sets the `class` of the div with `id=app` to `chat` or `standalone` depending on what the host app displays. This is bit hacky, but it demonstrates that this is possible and later can be improved. I added a more detailed descriptions to the caveats section about the issues with the current implementation.

Known issues:
- [x] Regarding the standalone components, I just realized that currently they render to the whole screen. I think it would be better if the header would be rendered by the host application, so that they can be always closed, and if needed navigation backwards would be allowed. I checked the design and for the implemented components there needs to be no change to achieve this. (https://github.com/logos-innovation-lab/waku-objects-playground/issues/368)
- [x] As mentioned above the standalone component currently uses the same JS code to render itself as the chat widget. It can differentiate between the states based on the `class` property of the top-level app `div`. This feels a bit sketchy, and discussing other alternatives would be appreciated. One idea I thought about but ultimately discarded was to inject the state in some variable with Javascript, but after #358 it seems to be better to use less Javascript than more of it. (https://github.com/logos-innovation-lab/waku-objects-playground/issues/369)
- [x] The iframe styling uses the default style, so it does not look like the app at all. I found some [ideas](https://stackoverflow.com/questions/4612374/iframe-inherit-from-parent) how you can copy the style between iframes, but I don't know how well it would work with the [luminance](https://www.npmjs.com/package/@waku-objects/luminance) design system, so that requires further research. (https://github.com/logos-innovation-lab/waku-objects-playground/issues/370)
- [x] The size of the iframe is now set based on the content of the iframe instead of the [default iframe size](https://stackoverflow.com/questions/5871668/default-width-height-of-an-iframe). For the chat widget the width should depend on the rest of the UI, because the chat widget width is a function of that (8/12 if I remember correctly). For standalone components it should fill the whole screen (minus the header). (https://github.com/logos-innovation-lab/waku-objects-playground/issues/370)
- [x] Currently all `DataMessage`s sent between the objects result in a rendered component. However there were discussions where sometimes there could be better UX if the `DataMessage` would not display a new bubble, but change the state of an existing one. With the current implementation there is no way to do it, and it would be great to cover this. For example the `WakuObjectDescriptor` could have a function that returns a `boolean` based on the message and the current state if the message needs to be displayed or not, or the `onMessage` could return a `boolean`. This is getting more interesting with the external object  because here we need to decide if we want to display a chat bubble or not before the iframe is displayed and logic runs in that. (https://github.com/logos-innovation-lab/waku-objects-playground/issues/371)

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -28,7 +28,6 @@ const config: KnipConfig = {
 	},
 	ignoreExportsUsedInFile: true,
 	ignoreBinaries: ['docker'],
-	ignoreDependencies: ['@waku-objects/sandbox-example'],
 }
 
 export default config
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
 		"@typescript-eslint/eslint-plugin": "^5.61.0",
 		"@typescript-eslint/parser": "^5.61.0",
 		"@waku-objects/luminance": "^2.0.1",
-		"@waku-objects/sandbox-example": "file:../waku-objects/objects/sandbox-example",
+		"@waku-objects/sandbox-example": "^0.4.0",
 		"@waku/interfaces": "^0.0.15",
 		"@waku/sdk": "^0.0.16",
 		"copy-to-clipboard": "^3.3.3",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
 		"@typescript-eslint/eslint-plugin": "^5.61.0",
 		"@typescript-eslint/parser": "^5.61.0",
 		"@waku-objects/luminance": "^2.0.1",
-		"@waku-objects/sandbox-example": "^0.3.0",
+		"@waku-objects/sandbox-example": "file:../waku-objects/objects/sandbox-example",
 		"@waku/interfaces": "^0.0.15",
 		"@waku/sdk": "^0.0.16",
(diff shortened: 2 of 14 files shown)


vojtechsimetka (approved):
LGTM

PR #382, opened 2023-09-13
feat: define split token in one place and use it everywhere in the app

### LINKED ISSUE #177: Refactor `Token` type so that it does not contain `amount` field

### LINKED ISSUE #379: Split: use native token symbol everywhere rather than hardcoded DAI

vojtechsimetka:
Resolves #177 
Resolves #379

### DIFF
--- a/src/lib/adapters/balance.ts
+++ b/src/lib/adapters/balance.ts
@@ -1,4 +1,4 @@
-import { balanceStore, type Token } from '$lib/stores/balances'
+import { balanceStore, type TokenAmount } from '$lib/stores/balances'
 import { defaultBlockchainNetwork, getBalance, getProvider } from '$lib/adapters/transaction'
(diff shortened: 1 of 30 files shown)


agazso (approved):
I think the naming would be better if there would be a `Token` type only defining the token. And there would be another type that would contain the `token` and the `amount`, just like you did. This could be called `TokenAmount` or just `Amount`.

vojtechsimetka:
> I think the naming would be better if there would be a `Token` type only defining the token. And there would be another type that would contain the `token` and the `amount`, just like you did. This could be called `TokenAmount` or just `Amount`.

Agree, will change :)

PR #385, opened 2023-09-13
feat: mentions

agazso:
This PR adds @mentions to group chats. It consists of several parts:
- Replaces the `@name` in the original text to `@address` in the actual text message before sending. This way if the users change their display name the mention will be still relevant.
- It replaces back to `@name` from `@address` before the message is displayed
- Any `@mention` will be replaced by **bold** text before the message is displayed.


vojtechsimetka (approved):
very very nice!

PR #389, opened 2023-09-14
feat: remove add images or object dropdown

vojtechsimetka:
This removes the add object/image dropdown. Instead, clicking the + button takes you directly to new object. In the last mile document we agreed that instead of uploading pictures we might implement galery waku object.

Screenshot 2023-09-14 at 12 01 08


agazso (approved):
This works, but I think it would be better to change the design so that it shows a dropdown with the Waku Objects. Therefore it would not need to navigate to a new page to select the Waku Object.

vojtechsimetka:
Agreed but I'd let David design it first.

PR #399, opened 2023-09-18
feat: estimate transaction cost

### LINKED ISSUE #314: Estimate transaction fee

vojtechsimetka:
resolves #314


src/lib/objects/split/views/settle-now.svelte
```diff
@@ -29,18 +29,31 @@
...
+	let owedAmount: undefined | bigint = undefined
 	let settling = false
-	$: {
-		const balance = balances.find(({ address }) => address === profile.address)
-		if (balance) {
-			owedAmount = BigInt(balance.amount)
-		}
-	}
+	let fee: bigint | undefined = undefined
+	let feeError: undefined | Error = undefined
```
agazso, line 35: `feeError` is not set anywhere. Maybe on line 51 it could be set?

src/lib/objects/split/blockchain.ts
```diff
@@ -1,17 +1,35 @@
...
-import splitterAbi from './abis/splitter.json'
+import { Interface, type Provider } from 'ethers'
+import type { GetContract } from './types'
+import splitterFactoryAbi from './contracts/abis/splitter-factory.json'
+import splitterAbi from './contracts/abis/splitter.json'
 import { defaultBlockchainNetwork } from '$lib/adapters/transaction'
 import type { Balance } from './schemas'
+import type { Splitter, SplitterFactory } from './contracts/types'
 
 const splitterFactoryAddress = defaultBlockchainNetwork.objects.splitterFactory
```
agazso, line 9: I just realized that the `splitterFactory` address is hardcoded in the `BlockchainNetwork` type. I don't think that's a good place to store it, it seems it would be better if the object code would define it, especially if the object is an external type then there is no way that it can expect the host application to know about it and expose it in the SDK.

This does not belong in this PR, but I just wanted to document this issue somewhere.

agazso (approved):
The code looks good, but please check out my comments.

I tested with Payggy where there is a field called `Transaction fee (max)` which looks reasonable. Then after the transaction is done the same is displayed but at that time the exact amount can be known and it would be better to display that instead of the estimated maximum which has no relevance anymore.

src/lib/objects/split/views/settle-now.svelte
vojtechsimetka: good catch! Forgot to commit this one, thanks!

src/lib/objects/split/blockchain.ts
vojtechsimetka: I kinda agree. At the same time, it also depends on which blockchain the app is connected to. I'll create an issue about this but I believe there is some discussion needed.

PR #400, opened 2023-09-18
feat: safe waku

agazso:
This PR adds a safe wrapper around sending and receiving waku messages. Also moves queueing and retry/resync logic to the safe wrapper from the adapter. The long term goal is to abstract away the connectivity related issues from the application.

Notes:
- There are known issues with `waku-js` (such as [this](https://github.com/waku-org/js-waku/issues/1569), [this](https://github.com/waku-org/js-waku/issues/1562) and [this](https://github.com/waku-org/js-waku/issues/1562)) that makes it very difficult or impossible to handle all errors gracefully in a wrapper at the moment.
- The Svelte `Readable` `subscribe` function calls the callback function immediately, which is not needed, that's why there are the `firstChatStoreSave` and `firstObjectStoreSave` variables.
- This safe wrapper approach does not work very well with the Svelte autoreload functionality. Editing and saving unrelated code may result in having multiple connection issues running in the background. In that case a refresh helps.
- Currently the logging is turned on with debug loglevel for the wrapper. This helps building intuition about the connectivity issues and also help to see when to refresh the browser (see the previous point).
- It would be good to disable the text inputs in the chat while the message is being sent and re-enable when it is sent. Currently only the send button is disabled but it is possible to send a message by pressing `Enter`. Maybe disabling sending with `Enter` can be a solution.
- Added preliminary support for multiple peers, however it looks like that if any of the connections are not reliable then it makes the initial connection more flaky. The solution could be to only block until there is at least one peer that is connected, but at the moment it's not clear what is the best way to achieve it. That will be addressed in a separate PR. (https://github.com/logos-innovation-lab/waku-objects-playground/issues/404)
- The disconnection flow is not heavily tested and there may be issues if there are queued messages. That will be addressed in a separate PR. (https://github.com/logos-innovation-lab/waku-objects-playground/issues/405)

### DIFF
--- a/src/lib/adapters/waku/safe-waku.ts
+++ b/src/lib/adapters/waku/safe-waku.ts
@@ -1,0 +1,232 @@
+import { connectWaku, type ConnectWakuOptions, sendMessage } from '$lib/adapters/waku/waku'
+import { isGroupChatId, type Message } from '$lib/stores/chat'
+import type { LightNode, TimeFilter, Unsubscribe } from '@waku/interfaces'
+import { PageDirection } from '@waku/interfaces'
+import { makeWakustore } from './wakustore'
+
+type Callback = (message: Message, chatId: string) => Promise<void>
+
+interface QueuedMessage {
+	chatMessage: Message
+	chatId: string
+	callback: Callback
+}
+
+interface Subscription {
+	unsubscribe: Unsubscribe
+	callback: Callback
+}
+
+export class SafeWaku {
+	public lightNode: LightNode | undefined = undefined
+	private subscriptions = new Map<string, Subscription>()
+	private lastMessages = new Map<string, Message>()
+	private lastSentTimestamp = 0
+	private isSubscribing = false
+	public readonly errors = {
+		numDisconnect: 0,
+		numSendError: 0,
+	}
+	private queuedMessages: QueuedMessage[] = []
+	private isHandlingMessage = false
+
+	constructor(public readonly options?: ConnectWakuOptions) {}
+
+	async connect() {
+		if (this.lightNode) {
+			return this.lightNode
+		}
+
+		this.lightNode = await this.connectWaku()
+		return this.lightNode
+	}
+
+	async subscribe(
+		chatId: string,
+		timeFilter: TimeFilter | undefined,
+		callback: (message: Message, chatId: string) => Promise<void>,
+	) {
+		if (!this.lightNode) {
+			this.lightNode = await this.connectWaku()
+		}
+
+		const lastMessageTime = this.lastMessages.get(chatId)?.timestamp || 0
+		const startTime = new Date(lastMessageTime + 1)
+		const endTime = new Date()
+		const calculatedTimeFilter = lastMessageTime
+			? { startTime, endTime }
+			: { startTime: endTime, endTime }
+		timeFilter = timeFilter || calculatedTimeFilter
+
+		const talkEmoji = isGroupChatId(chatId) ? '🗫' : '🗩'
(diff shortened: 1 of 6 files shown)


src/lib/adapters/waku/safe-waku.ts
```diff
@@ -0,0 +1,243 @@
...
+
+		const oldSubscriptions = this.subscriptions
+		this.subscriptions = new Map()
+
+		for (const chatId of chatIds) {
+			const callback = oldSubscriptions.get(chatId)?.callback
+			if (callback) {
+				await this.subscribe(chatId, undefined, callback)
+			}
+		}
```
filoozom, line 189: Maybe it would be good to `try`/`finally` this to avoid errors that could make it so that `this.isSubscribing = false` is never reached, even though some of these functions already have `try`/`catch`.

src/lib/adapters/waku/safe-waku.ts
```diff
@@ -0,0 +1,243 @@
...
+					lastMessage.type === message.type &&
+					lastMessage.fromAddress === message.fromAddress
+				) {
+					this.log('🙈 ignoring duplicate message', { message, lastMessage })
+					continue
+				}
+
+				this.lastMessages.set(chatId, message)
+
+				await callback(queuedMessage.chatMessage, queuedMessage.chatId)
```
filoozom, line 225: This could technically also throw and thus cancel `this.isHandlingMessage = false`.

src/lib/adapters/waku/waku.ts
```diff
@@ -57,7 +61,10 @@ export async function connectWaku(options?: ConnectWakuOptions) {
 	})
 
 	await waku.start()
-	await waku.dial(peerMultiaddr)
+	for (const peer of peers) {
+		const addr = multiaddr(peer)
+		await waku.dial(addr)
+	}
```
filoozom, line 67: What about something like:
```suggestion
	await Promise.race(peers.map(peer => waku.dial(multiaddr(peer)))
```
This way it will continue execution after one peer is connected, but still try to connect to the other ones in the background.

filoozom (approved):

src/lib/adapters/waku/safe-waku.ts
agazso: Yes, that's a good point, I will wrap this.

src/lib/adapters/waku/safe-waku.ts
agazso: On a second thought I am not sure the `isSubscribing` is really needed here. It would be better to make sure that the `safeSubscribe` is not called in parallel.

But you bring up a valid point that if the `subscribe` throws an exception currently it causes an unrecoverable error. I will see what can be done about it.

src/lib/adapters/waku/waku.ts
agazso: I thought of that but then there is also an async call to `waitForRemotePeer` after that. I am not sure how the combination of these things work and need to further look into their source code.

Another problem is that this may cause an unhandled rejection if the `dial` of a non-winning peer fails for some reason. I am trying to be more mindful of these issues after experiencing similar bugs from js-waku.

agazso:
@filoozom fixed the `resubscribe` and the `callback`, please re-review it.

Also added a safe wrapper around `connect`.

src/lib/adapters/waku/waku.ts
agazso: https://github.com/logos-innovation-lab/waku-objects-playground/issues/404

PR #422, opened 2023-09-20
feat: dicebear icons

### LINKED ISSUE #392: default avatar with dicebear

agazso:
Using the `bottts-neutral` style for users and the `shapes` style for groups.

Closes #392


src/lib/components/avatar.svelte
```diff
@@ -1,11 +1,17 @@
...
 	export let size = 48
 	export let onClick: (() => unknown) | undefined = undefined
-	export let avatarSize = size / 2
+	export let group = false
+
+	function isCID(s: string) {
+		return s.length === 46 && s.startsWith('Qm')
+	}
+
+	console.debug('avatar', { picture, group })
```
vojtechsimetka, line 14:
```suggestion
```

src/lib/components/avatar.svelte
```diff
@@ -16,10 +22,11 @@
 	tabindex={0}
 >
 	<div class="img" style={`height: ${size}px;`}>
-		{#if picture}
+		{#if picture && isCID(picture)}
```
vojtechsimetka, line 25: Personally I would use another prop for the so that we don't need to check for CID. So if the Avatar has `picture` defined, show that if not, show `dicebear` with `dicebearSeed` prop. So we would use it as `<Avatar picture={chat.picture} dicebearSeed={chat.id} />

But I am fine with this implementation as well, just wanted to check if you considered this.

vojtechsimetka (approved):

src/lib/components/avatar.svelte
agazso: Something is missing from your comment and I don't understand what exactly you are saying here :smile:

vojtechsimetka: Ah weird... Anyways the gist was if you considered to not pass the `chat.id` for dicebear as `picture` but instead as separat props. This way there is no need to check if it `isCID` and it serves as fallback.

`<Avatar picture={chat.picture} dicebearSeed={chat.id}`

agazso: Okay, I added a `seed` prop because the code is better without the conditionals everywhere.

src/lib/components/avatar.svelte
```diff
@@ -20,11 +17,11 @@
...
 	<div class="img" style={`height: ${size}px;`}>
-		{#if picture && isCID(picture)}
+		{#if picture}
 			<img src={getPicture(picture)} alt="profile" />
-		{:else if picture}
+		{:else if seed}
 			{@const style = group ? 'shapes' : 'bottts-neutral'}
-			<Dicebear name={picture || ''} {size} {style} />
+			<Dicebear {seed} {size} {style} />
 		{/if}
```
vojtechsimetka, line 25: Can it happen that both `picture` and `seed` is undefined? I guess then we only see grey circle, right? I can live with that.

vojtechsimetka (approved):

src/lib/components/avatar.svelte
agazso: Normally this cannot happen, because there is always an address to be used for seed. But yes if this is the case then the result is a gray circle.

PR #426, opened 2023-09-21
feat: chat list timestamp

### LINKED ISSUE #25: Timestamps in chat lists

agazso:
Closes #25


vojtechsimetka (approved):
Do we still need the `formatDateAndTime` function or should we use everywhere the new `formatTimestamp`?

agazso:
> Do we still need the `formatDateAndTime` function or should we use everywhere the new `formatTimestamp`?

It depends on the context. I am now working on the chat timestamps and they require two more (similar but different) implementations of a time formatting function.

Issue #347, opened 2023-09-06
Missing avatar and name for group chat WO bubbles


agazso:
Currently there are no avatars and names displayed next to Waku Object chat bubbles in a group chat. 

Here is the reference design: https://www.sketch.com/s/755dadbd-2e1d-469f-bd5d-de887a0543cb/a/1V4E0Jp

PR #428, opened 2023-09-21
feat: chat timestamp

### LINKED ISSUE #32: Timestamps and day indicator in messages

agazso:
Closes #32 

There were several simplifications compared to the original design: 
- [ ] In the design the timestamps are on the same line if the message is only one line
- [ ] The badges indicating the days are floating in the design, I just implemented them as markers between the chat bubbles.

Screenshot from 2023-09-21 18-18-59

### DIFF
--- a/src/lib/components/chat-date-badge.svelte
+++ b/src/lib/components/chat-date-badge.svelte
@@ -1,0 +1,24 @@
+<script lang="ts">
+	export let text: string
+</script>
+
+<div class="time-separator">{text}</div>
+
+<style lang="scss">
+	.time-separator {
+		background-color: var(--color-step-20, var(--color-dark-step-40));
+		color: var(--color-step-50, var(--color-dark-step-20));
+		font-size: var(--font-size-sm);
+		border-radius: var(--border-radius);
+		text-align: center;
+		margin-left: auto;
+		margin-right: auto;
+		margin-top: var(--spacing-24);
+		margin-bottom: var(--spacing-24);
+		max-width: fit-content;
+		padding-left: var(--spacing-6);
+		padding-right: var(--spacing-6);
+		padding-top: var(--spacing-3);
+		padding-bottom: var(--spacing-3);
+	}
+</style>
--- a/src/lib/components/chat-message.svelte
+++ b/src/lib/components/chat-message.svelte
@@ -18,6 +18,8 @@
 	//name of the person who sent the message
 	export let senderName: undefined | string = undefined
 
+	export let timestamp: string | undefined = undefined
+
 	const isFF = () => {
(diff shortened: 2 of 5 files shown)


vojtechsimetka (approved):
So this is definitely an improvement, but it does not play nice with object. We can tackle that separately, just FYI

agazso:
> So this is definitely an improvement, but it does not play nice with object. We can tackle that separately, just FYI

What is the issue? Maybe I can address it in this PR.

vojtechsimetka:
Yes to few things:

1. Timestamps don't exist on waku object messages
2. In the following screenshot, I've sent the waku object today and then the text. However, the `Today` only appeared after the waku object and before the normal text. There is around 1 min difference between these two.
Screenshot 2023-09-22 at 18 01 32

3. If I scroll all the way to the top, it shows just time and I have no idea when the message was sent (it was sent 2 days ago). So before the first message in the chat we should always have the date.
Screenshot 2023-09-22 at 18 01 38

agazso:
>     1. Timestamps don't exist on waku object messages

This is an existing issue and I didn't necessarily want to fix in this PR: https://github.com/logos-innovation-lab/waku-objects-playground/issues/347

>     2. In the following screenshot, I've sent the waku object today and then the text. However, the `Today` only appeared after the waku object and before the normal text. There is around 1 min difference between these two.

This is indeed an issue, and I will fix it.

>     3. If I scroll all the way to the top, it shows just time and I have no idea when the message was sent (it was sent 2 days ago). So before the first message in the chat we should always have the date.

This is a new feature request and I think it's a good idea and seems easy to add, so I will add it too.

agazso:
I added a fix that addresses points 2 and 3. 

As for the first one, currently all the objects render their own `ChatMessage` components and that's where the timestamp needs to be passed from the message (also the name and avatar in #347). It would be better if the host app would render the `ChatMessage` frame with the designed embellishments and the app would just render inside a rectangle. I talked to @david-gauquelin about this and he plans to document it as part of the design system. Then we can refactor the code for the objects to work that way.

michelleplur:
Looks really good ❤️

vojtechsimetka (approved):

david-gauquelin:
ok so answering all the different points in the thread here:

- yes I'm defining in the guidelines what belongs to the chat and what belongs to a Waku Object (in chat bubble), but definitely the time stamp belongs to the chat app so it should appear on Waku Object messages in-chat
- the floating day/date (always visible) would be really nice as otherwise I just see the time and it can be confusing/misleading
- the time on one line is a nice to have, having on 2 lines is less space-efficient but doesn't create usability issues
- finally, please use a sans-serif font for the timestamp, the serif font is for the message text only

PR #429, opened 2023-09-21
fix: split object works after adding new members to chat

### LINKED ISSUE #401: Existing split object will not allow adding more expenses after someone new is added to the chat
The reason for this is that with each new expense we define who are the members that share that expense. Right now we use all the members of the chat. However, the splitter contract checks if these expense members are also members of the splitter collection and if someone is not a member, reverts the transaction.

The best solution is to not specify members in each expense but this is not supported by contracts. So the next best thing is to take the members from the object store.

vojtechsimetka:
resolves #401


michelleplur:
Nice

PR #442, opened 2023-09-25
feat: heic image support

### LINKED ISSUE #430: Missing error when uploading HEIC profile picture
When creating a new identity, I tried to upload a HEIC image as a profile. It was not supported, nothing happened but I should have an error/info about it

vojtechsimetka:
Resolves #430 

We have to lazy-load the library as it only works in browser. That's why there is:
```ts
	const heic2any = await import('heic2any')
	const out = await heic2any.default({
		blob: heicFile,
		toType: 'image/png',
	})
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
 		"eslint-plugin-svelte3": "^4.0.0",
 		"ethers": "^6.6.4",
 		"hardhat": "^2.16.1",
+		"heic2any": "^0.0.4",
 		"html5-qrcode": "^2.3.8",
 		"ipfs-http-client": "^60.0.0",
 		"knip": "^2.17.1",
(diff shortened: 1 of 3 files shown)


vojtechsimetka:
It works now even for that image. Did not know these can also be `image/heif` I thought just `image/heic`

agazso:
> It works now even for that image. Did not know these can also be `image/heif` I thought just `image/heic`

I don't get an error now, but nothing happens when I try to set the linked HEIC image as my profile picture. Does it work for you?

michelleplur:
does it also scale down the image to a smaller filesize?

vojtechsimetka:
Yeah it scales them down and it may take quite some time. We should show some loading. The linked `heic` works for me 
Screenshot 2023-09-25 at 22 09 54

agazso:
Well, it does not work for me neither in Firefox or Chrome. But I guess if it works on iOS Safari then it's okay.

vojtechsimetka:
It works in chrome for me

vojtechsimetka:
@agazso please try again now, I read up a bit on heic/heif and I missed on condition checking `heif` type

agazso (approved):
It works!

PR #444, opened 2023-09-25
docs: chat protocol

### LINKED ISSUE #390: write up DM protocol

### LINKED ISSUE #391: write up group chat protocol

agazso:
This is the first part of the docs that is about the chat protocol and related type definitions.

Closes #390
Closes #391


DEVELOPERS.md
```diff
@@ -0,0 +1,159 @@
...
+  - Types
+  - External objects
+  - Creating your own object type
+  - Existing implementations
+
+## Overview
+
+Waku Objects is an open, modular system for transactional mini-applications to be built inside a chat application. Open, because it uses open standards for interoperability and provides a default [web based implementation](https://www.wakuplay.im/). Modular, because it may be implemented in any chat application that supports structured data messages that are sent alongside user messages. Transactional, because it supports blockchain transactions and smart contract interactions on Ethereum compatible blockchains.
+
+The Waku Object concept enables building social applications, where the social context is provided by the chat (private 1on1 or group at the moment) so that there is already some implied trust between the participants. The Waku Object SDK provides an interface for objects to communicate easily with the other participants' objects and offer safe services without leaking private or personal data.
```
vojtechsimetka, line 28: > ...and offer safe services without leaking private or personal data

I am not sure we can claim this right now.

vojtechsimetka (approved):
Very very nice!

DEVELOPERS.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+There is a field `type` that is `invite`, the `timestamp` is the number of milliseconds since the Unix Epoch in UTC. The `fromAddress` is the ethereum address of the sender as a string, prefixed with `0x`. The `objectId` is the unique id of the Waku Object, for external objects it is the same as their `npm` package name. Each object may have multiple instances. The `instanceId` is a unique identifier, in the current implementation it is a 12-characters long random hexstring. The `data` field contains the object specific data, the only requirement for it is it have to be JSON serializable, so there can be no circular references in it.
+
+### Private chat
+
+The private chat is defined as a private conversation between two participants. Each user has a pubsub topic derived from their Ethereum address, where others can send messages. The topic is defined as follows:
+
+```typescript
+;`/${topicApp}/${topicVersion}/${contentTopic}/${id}`
```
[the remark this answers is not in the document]
agazso: It seems that `pnpm format` added it there. I will change this to something that will work with `pnpm format`.

DEVELOPERS.md
agazso: I reworded it to sound more aspirational than factual. 

What I meant to say here is that the Waku Object interface is designed to share the least information possible in a sandboxed environment so that the possibility of leaking private or personal data is minimised. We don't share messages, we don't share the private key, the only thing we share currently is the name, avatars and addresses of the participant. From these the names and avatars are arbitrary, so it's only the addresses that can be considered PII. We can even use generated addresses and replace them in the application so that the objects has not even knowledge of the real addresses. I will open an issue about this :smile:

agazso: https://github.com/logos-innovation-lab/waku-objects-playground/issues/448

Issue #450, opened 2023-09-26
splitterFactory address is hardcoded in the BlockchainNetwork type


agazso:
> I just realized that the `splitterFactory` address is hardcoded in the `BlockchainNetwork` type. I don't think that's a good place to store it, it seems it would be better if the object code would define it, especially if the object is an external type then there is no way that it can expect the host application to know about it and expose it in the SDK.

_Originally posted by @agazso in https://github.com/logos-innovation-lab/waku-objects-playground/pull/399#discussion_r1330099175_

vojtechsimetka:
But if we do this, then we need to expose `chainId` to the object

agazso:
Yeah, let's do that! I would put this in the `WakuObjectState`

PR #451, opened 2023-09-26
feat: use gnosis chain by default

### LINKED ISSUE #449: Use Gnosis chain by default

agazso:
- [x] Change `splitterFactory` address (but also see #450)

Closes #449

### DIFF
--- a/src/lib/adapters/transaction.ts
+++ b/src/lib/adapters/transaction.ts
@@ -93,7 +93,7 @@ const gnosisBlockchain: BlockchainNetwork = {
 		},
 	],
 	objects: {
-		splitterFactory: '0x941DDB22a33FC753d3E7b82cc34c47Ee605e60a3',
+		splitterFactory: '0xCdE164E4274Ee30dA21cf5407B8FB37A265f623C',
 	},
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/adapters/transaction.ts
```diff
@@ -88,7 +116,7 @@ const sepoliaBlockchain: BlockchainNetwork = {
 	},
 }
 
-export const defaultBlockchainNetwork = chiadoBlockchain
+export const defaultBlockchainNetwork = gnosisBlockchain
```
vojtechsimetka, line 119: Can we make this env variable? So something like
```suggestion
export const defaultBlockchainNetwork = blockchains[PUBLIC_ENV_CHAIN_NAME] ?? testnet
```

agazso: I don't mind making this configurable with an env variable, but I would make the gnosis the default, otherwise you must configure it everywhere where we test (locally, on vercel, etc.)

agazso: Not going to make this change in this PR, but I opened an issue about it: https://github.com/logos-innovation-lab/waku-objects-playground/issues/452

src/lib/adapters/transaction.ts
```diff
@@ -69,6 +69,34 @@ const chiadoBlockchain: BlockchainNetwork = {
 	},
 }
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const gnosisBlockchain: BlockchainNetwork = {
```
filoozom, line 73: Maybe `export` this and add an optional argument to `getProvider`?

filoozom (approved):

Issue #360, opened 2023-09-08
Waiting animations and spinner


agazso:
There are many times a user interaction has to wait several seconds (e.g. blockchain transactions). It would be good to have a general guideline how to add animations to buttons, overlays or transitions. A default spinner component would be nice too.

david-gauquelin:
I suggest that we use the 2nd example from here https://tobiasahlin.com/spinkit/ 

either:

- centered on a page (if loading a new page)
- inside a button shape (replacing its content)
- or in an alert (see the Loading and errors section in "Waku Chat" Sketch file)

vojtechsimetka:
I don't see an example of the loading in sketch, maybe you did not publish it yet. But I assume when it comes to button, it replaces the icon, right?
![Screenshot 2023-09-25 at 15 34 49](https://github.com/logos-innovation-lab/waku-objects-playground/assets/7974813/4ec27257-8d79-41d5-b447-bb721427862b)

david-gauquelin:
I didn't do a mockup for it, I meant to look at what the alert looks like in Sketch.
Basically just put that spinner centered in the container (page, alert or button).

In the case of a button you can either replace just the icon, or the whole button content (whatever is easier)

PR #453, opened 2023-09-28
feat: loading component

vojtechsimetka:
Partially addresses #360 , I will go through the app and review where we need to have loading state. Also implement it in buttons.


src/lib/components/loading.svelte
```diff
@@ -0,0 +1,98 @@
+<script lang="ts">
```
agazso, line 1: I think an attribution to the creator would be nice: http://tobiasahlin.com/spinkit/

agazso (approved):

PR #455, opened 2023-09-28
feat: waku multiconnection

agazso:
Add more waku nodes to connect to besides the existing one to improve reliability


agazso:
This is done better in #466

PR #472, opened 2023-10-12
feat: extract splitterFactory address to the object and expose chain ID

### LINKED ISSUE #450: splitterFactory address is hardcoded in the BlockchainNetwork type

vojtechsimetka:
Resolves #450

I decided to stor the `chainId` of the different chains into the `BlockchainNetwork` type because:

1. Getting the `chainId` is async call which makes the `args` creation a bit more challenging. Storing it in the `BlockchainNetwork` means we can just get it.
2. This way we can define the chains we want to connect to and then validate with the RPC that we are connected to the right chain. This is done in the main `Layout`.

### DIFF
--- a/src/lib/adapters/transaction.ts
+++ b/src/lib/adapters/transaction.ts
@@ -18,17 +18,16 @@ interface BlockchainExplorer {
 
 interface BlockchainNetwork {
 	name: string
+	chainId: bigint
 	provider: string
 	explorer?: BlockchainExplorer
 	nativeToken: Token
 	tokens?: Token[]
-	objects: {
-		splitterFactory: string
-	}
 }
 
 const localBlockchain: BlockchainNetwork = {
 	name: 'Local testnet',
+	chainId: 31337n,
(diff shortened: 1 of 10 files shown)


agazso (approved):
Looks good.

I found that the `chainId` is `bigint` strange but I reckon it is also returned as such in Ethers.

vojtechsimetka:
> I found that the `chainId` is `bigint` strange but I reckon it is also returned as such in Ethers.

Yes `ethersjs@6` does return everything as `bigint` now. Token balances, or even chain IDs. Before these used to be `strings`

PR #474, opened 2023-10-19
feat: installed object list

agazso:


vojtechsimetka:
I don't quite understand what this does. Can you explain a bit why we need the notion of installed objects? I assume this is some preparation for installing the objects from the chat but not quite sure what is the purpose right now (if any).

agazso:
> I don't quite understand what this does. Can you explain a bit why we need the notion of installed objects? I assume this is some preparation for installing the objects from the chat but not quite sure what is the purpose right now (if any).

Yes, this is the preparation for installable objects. Each user may have their own set of installed objects and it needs to be stored somewhere. Currently we are storing user data in Svelte stores, so this PR creates a new one for installable objects.

vojtechsimetka (approved):

PR #483, opened 2023-10-27
feat: sets the right theme color (for iPhones)

filoozom:
This should probably be refactored into a:
```html
<svelte:head>
	<meta name="theme-color" content={$theme.isDark ? 'black' : 'white'} />
</svelte:head>
```

The issue is that currently there's no easy way to listen to changed about which theme is actually being displayed. If it's the system theme and the system appearance changes, the store does not. The `isSystemDark` variable and listener from #482 should probably be moved to the store?

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -7,7 +7,6 @@
 		<link rel="apple-touch-icon" href="/images/icons/icon-192x192.png" />
 		<link rel="manifest" href="/manifest.json" />
 		<meta name="msapplication-TileColor" content="#da532c" />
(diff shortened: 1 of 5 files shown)


filoozom:
Alright, I put everything in the store. I debated using a `derived([$theme, $isSystemDark])`, but that would then require the use of two stores for the base color and whether or not the dark theme is activated, so I think this way is a bit cleaner.

vojtechsimetka (approved):

PR #489, opened 2023-11-07
feat: external objects with URL and metadata check

filoozom:
Needs some rebasing and doesn't include the chat flow to add objects. Maybe some additional checks for IPFS should be added as well, but they would be redundant assuming we agree on the `metadata.json` format.

There's a missing piece that is checking the `metadata.json` signature, but that should also be done in the chat flow.

This PR also fixes the CSP security issue where the entrypoint needed to be specified with a URL rather than a hash. The fix was to add the `integrity` prop to the `<script>` tag. I haven't tried this on Firefox though, hopefully this didn't introduce an additional CSP bug.

An example object to be added through the form on the settings tab: `url:https://unpkg.com/@waku-objects/sandbox-example@0.5.0/object`.

### DIFF
--- a/src/lib/objects/external/iframe.svelte
+++ b/src/lib/objects/external/iframe.svelte
@@ -7,7 +7,7 @@
 	import template from './template.html?raw'
 
 	// Types
-	import { getNPMObject, type LoadedObject } from './lib'
+	import { getObject, type LoadedObject } from './lib'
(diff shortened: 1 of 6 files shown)


agazso (approved):
Did not test it, because I am not exactly sure how, but the code looks good.

PR #494, opened 2023-11-09
feat: encryption with signatures

agazso:
**This is a breaking change**

Fixes some of the issues created by this PR: https://github.com/logos-innovation-lab/waku-objects-playground/pull/488
Basically this PR added signing and verification to the data and also changed the code to work with encrypted encoders/decoders as well.
Originally it also contained ECIES encryption but that does not work at the moment in js-waku: https://github.com/waku-org/js-waku/issues/1716


vojtechsimetka (approved):
tested and works good

PR #495, opened 2023-11-10
feat: babbles logic proof-of-concept

agazso:
This PR implements the proof-of-concept logic and protocol for the babbles feature (but not the UI as it is not yet specified completely).

- Creation of babbles with a random topic, name and avatar (maybe not needed). 
- New message type defined for babbles with the option to add children messages (implemented with a `parentId` pointer). Each message has an implicit id which is the [message hash](https://rfc.vac.dev/spec/14/#deterministic-message-hashing).
- There is a function for converting the list of messages to a threaded view (`convertMessagesToThreaded`). This can be  extended according to the required features (sorting, hiding etc.) later. 
- As a result of a refactor the message decoding logic moved from `wakustore` to `safe-waku` and there is a separation of `Message`s that are sent and the internal representation (`ChatMessage`). The latter includes `Metadata` as well, for that a new type has been introduced.

### DIFF
--- a/src/lib/adapters/waku/safe-waku.ts
+++ b/src/lib/adapters/waku/safe-waku.ts
@@ -31,7 +31,7 @@ async function sleep(msec: number) {
 export class SafeWaku {
 	public lightNode: LightNode | undefined = undefined
 	private subscriptions = new Map<string, Subscription>()
-	private lastMessages = new Map<string, Message>()
+	private lastMessages = new Map<string, ChatMessage>()
 	private isReSubscribing = false
 	public readonly errors = {
 		numDisconnect: 0,
@@ -290,17 +290,14 @@ export class SafeWaku {
 		this.subscriptions.set(chatId, subscription)
 	}
 
-	private areMessagesEqual(a: Message, b: Message): boolean {
+	private areMessagesEqual(a: ChatMessage, b: ChatMessage): boolean {
 		if (a.timestamp !== b.timestamp) {
 			return false
 		}
-		if (a.type !== b.type) {
+		if (a.senderPublicKey !== b.senderPublicKey) {
 			return false
 		}
-		if (a.type === 'babble' && b.type === 'babble' && a.id !== b.id) {
-			return false
-		}
-		if (a.type !== 'babble' && b.type !== 'babble' && a.senderPublicKey !== b.senderPublicKey) {
+		if (a.id !== b.id) {
 			return false
 		}
 
--- a/src/lib/adapters/waku/safe-waku.ts
+++ b/src/lib/adapters/waku/safe-waku.ts
@@ -1,15 +1,17 @@
 import { connectWaku, storeDocument, type ConnectWakuOptions } from '$lib/adapters/waku/waku'
-import type { Message } from '$lib/stores/chat'
+import type { ChatMessage, Message } from '$lib/stores/chat'
 import type { IDecoder, IEncoder, LightNode, TimeFilter, Unsubscribe } from '@waku/interfaces'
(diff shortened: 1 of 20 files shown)


src/lib/adapters/waku/safe-waku.ts
```diff
@@ -104,7 +106,7 @@ export class SafeWaku {
 		this.subscriptions = new Map()
 	}
 
-	async sendMessage(message: Message, encryptionKey: Uint8Array, sigPrivKey: Uint8Array) {
+	async sendMessage(message: unknown, encryptionKey: Uint8Array, sigPrivKey: Uint8Array) {
```
vojtechsimetka, line 109: Why is the message `unknown` now? I would assume that when sending we know the shape, when receiving we don't.

src/lib/adapters/waku/safe-waku.ts
```diff
@@ -275,10 +286,27 @@ export class SafeWaku {
...
+		if (a.type === 'babble' && b.type === 'babble' && a.id !== b.id) {
+			return false
+		}
+		if (a.type !== 'babble' && b.type !== 'babble' && a.senderPublicKey !== b.senderPublicKey) {
+			return false
+		}
+
+		return true
+	}
+
```
vojtechsimetka, line 305: This is to avoid including mesages twice? Why don't we check the content as well?

src/lib/components/chat-message.svelte
```diff
@@ -20,16 +20,24 @@
...
+	export let leftPadding: number | undefined = undefined
+
 	const isFF = () => {
 		let browserInfo = navigator.userAgent
 		return browserInfo.includes('Firefox')
 	}
 </script>
 
+<!-- svelte-ignore a11y-no-static-element-interactions -->
+<!-- svelte-ignore a11y-click-events-have-key-events -->
```
vojtechsimetka, line 34: note for self: this sould be resolved and removed

vojtechsimetka (approved):

src/lib/adapters/waku/safe-waku.ts
agazso: Good question! :smile: 

Historically this was originally in the `waku.ts` but I lifted it here, because it had an identical implementation as the `storeDocument` function. Now there is only `storeDocument` at the waku module and this one just calls the `storeEncrypted` which is just a safe wrapper around `storeDocument`.

You are right that here we know the shape, however I did another refactor on the `Message` type. We differentiate between the messages that we send and the messages that we receive. The former does not have metadata and that is what is sent over the wire, the latter has because we augment it with the metadata that we receive from the waku `DecodedMessage` type.  

So this should be a message without metadata type, however I did not manage to express that correctly in Typescript. I just added a commit to indicate that this has the type `WithoutMeta<Message>` but Typescript narrows it to just `{ type: 'user' | 'data' | 'invite' | 'babble' }` which does not provide real type safety. However it is better than `unknown` and at least expresses the intention.

src/lib/adapters/waku/safe-waku.ts
agazso: Good catch! Now that we have metadata it is enough to check the equality for `timestamp`, `id` and `senderPublicKey`. I added a fix for this.

src/lib/components/chat-message.svelte
agazso: I just added a comment for this: `TODO: remove onClick and the ignores once the real design is implemented for babbles`

This is just a temporary hack so that you can open the chat messages, but I think once we have a real design for babbles we will use a different component and then the `onClick` propery can be removed.

PR #497, opened 2023-11-21
feat: install invite flow

agazso:
This PR implements the invite flow for installing objects in a chat. It works in private chats and group chats. Only not preinstalled objects can be invited to install. Preinstalled objects (that are part of the app) will simply create an instance, but this may break if the other party does not have the same set of preinstalled objects. Because of this the best would be to not have preinstalled objects at all, but instead have downloadable objects installed as part of the app, therefore they can be shared as well.

There is also the [waku object list](https://www.sketch.com/s/4dc77780-04fa-4581-968f-4b78c0a2e737/a/Gon7LdO) implemented which can be accessed from the Identity page.

Known issues:
- [ ] The decline flow is missing. It would require a new message type (or `command` to the `InstallMessage` type) and it should reset the state for the inviter (but only in private chat, not in group). 
- [ ] Multiple invites can be sent. Not a huge issue but it is not very nice. The fix would require to track the invite state per chat.
- [ ] The number of invites in a group is missing too. The fix would require to track the number of invites and their state per chat.
- [ ] There is an uninstall button on the waku object page for downloadable objects. Currently this button removes the object from the downloadable object list, but it does not remove the object from chats. This can be used for development purposes, but this may break if an downloadable object was already used in a chat. The solution could be to add a new flag to indicate that the object was removed and change the behavior based on that.
- [ ] The downloadable objects currently assume that they are either `chat` apps or `standalone` apps and the `getObjectSpec` function expects this to be specified before downloading the metadata. This is not correct because most apps may have both and this should be handled differently (by checking the `class` property of the element named `app` or `document.getElementById('app').getAttribute('class')`). Because of this currently only `chat` apps are supported.

Instructions for testing:
- You need a downloadable object first to be able to test the complete flow. For that go to `Identity` - `Waku Objects`, paste this link in the `Object path` and press `Add object`: `url:https://unpkg.com/@waku-objects/sandbox-example@0.5.0/object`
- It will add a new object called `Sandbox example`. Please note that this does not have an asterisk* after the name, this indicates that it is downloadable.
- Go to a chat and press `+` to add a new object. Choose the one without the asterisk* and press `Send invite`
- Then the other user has to `Accept` the invite
- After that the object can be used in the chat


vojtechsimetka (approved):
All tested, works good.

PR #504, opened 2023-12-04
fix: babbles messages

agazso:
Sometimes messages disappear from the waku store. This is a fix for the case if a message's parent is lost to keep at least the rest of the thread.

### DIFF
--- a/src/routes/babbles/chat/[id]/[[thread_id]]/+page.svelte
+++ b/src/routes/babbles/chat/[id]/[[thread_id]]/+page.svelte
@@ -38,6 +38,8 @@
 	$: chatMessages = $chats.chats.get($page.params.id)?.messages || []
 	$: displayMessages = convertMessagesToThreaded(chatMessages)
 
+	$: console.debug({ chatMessages, displayMessages })
+
 	beforeUpdate(() => {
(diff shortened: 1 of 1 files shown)


src/routes/babbles/chat/[id]/[[thread_id]]/+page.svelte
```diff
@@ -126,6 +126,9 @@
 
 			const parentMessage = parentMap.get(message.parentId)
 			if (!parentMessage) {
+				// this is a workaround for the case when the parent is lost to
+				// keep the rest of the thread
+				roots.push(parent)
```
Comment (vojtechsimetka) [inline src/routes/babbles/chat/[id]/[[thread_id]]/+page.svelte:131]: Is this a temporary fix? If so, I suggest adding `//FIXME:` or `//TODO:`

vojtechsimetka (approved):

src/routes/babbles/chat/[id]/[[thread_id]]/+page.svelte
Comment (agazso) [reply src/routes/babbles/chat/[id]/[[thread_id]]/+page.svelte:131]: Well, for now it is a workaround for lost messages. 

However there is a broader issue that we only keep MAX_MESSAGES (currently set to 100) messages locally in the history. This is not a big issue with chats because they are linear, but with babbles it is possible that a parent message is lost because of this and the conversation would be orphaned without this workaround too.

PR #506, opened 2023-12-05
fix: message queue

agazso:
Change `queueMessage` to not block (also change the callback type to be async).

### DIFF
--- a/src/lib/adapters/waku/wakustore.ts
+++ b/src/lib/adapters/waku/wakustore.ts
@@ -24,7 +24,7 @@ export interface Wakustore {
 	collectionQuery: (decoder: IDecoder<DecodedMessage>, queryOptions?: QueryOptions) => Query
 	onSnapshot: <T>(
 		query: Query,
-		callback: (value: T, decodedMessage: DecodedMessage) => void,
+		callback: (value: T, decodedMessage: DecodedMessage) => void | Promise<void>,
 	) => Promise<Unsubscribe>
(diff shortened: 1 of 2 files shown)


src/lib/adapters/waku/wakustore.ts
```diff
@@ -115,23 +115,24 @@ export function makeWakustore(waku: LightNode): Wakustore {
...
 			pageSize: query.queryOptions?.pageSize ?? query.queryOptions?.limit,
 		}
+
 		const result = await readStore(waku, query.decoder, queryOptions)
 		const messages = await getQueryResults(result, queryOptions)
 
 		for (const message of messages) {
 			const value = decodeDoc<T>(message)
-			callback(value, message)
+			await callback(value, message)
```
vojtechsimetka, line 135: I wonder if we need to have some validation here. What if the message is not of type `T`?

vojtechsimetka (approved):

src/lib/adapters/waku/wakustore.ts
agazso: Currently there is no validation at all, so there can be two different types of error:
1. If the message payload is not JSON then `decodeDoc` will fail with an error.
2. If the message is JSON but the type is not `T` then most likely the `callback` will fail.

It would be nice to have validation here and because this is a generic function the solution could be to pass in a generic (optional?) validator function that fails early.

vojtechsimetka: Yeah I did that approach in the read from localstore helper fuction. It takes optional schema which it validates against.

